PR #1, opened 2020-12-17
Initial repo setup

dehall:
Initial setup of the repo and project structure. Some items of note:

 - This is inspired by the `express-generator` npx tool. "Inspired by" because I setup the repo before knowing that existed, so when I found it I merged things over. Also it doesn't use a `src` directory which I think we should.
 -  We can enable logging by setting the `DEBUG` flag in `.env`, for example `DEBUG=*` will include all app/server logs, or you can filter by specific tags
 - Default port is 3000, can be overridden in `.env`
 - The api and network aspect of the project are split out into separate files. `bin/www` is a script that sets up the http config (host/port/etc) and `src/app.js` is where the api is defined. This makes testing the api a lot easier. See https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/projectstructre/separateexpress.md for a little more info


KeeyanGhoreshi (approved):
Looks good, everything worked as expected.

dubem7 (approved):
Looks good to me

PR #5, opened 2021-01-21
Fetch KnowledgeArtifacts on Startup

blangley28:
When the server is spinning up it retrieves all PlanDefinitions (and references resources) from all Knowledge Artifact Repositories listed in the config file. I added the IG example to the KA running on pathways.mitre.org
The two questions I have about this implementation:
1. Is using `_include=*` suffice? I suppose not every FHIR server will necessarily support this. 
2. All the resources are stored in a collection named "{resource_type}s" (all lower case). An alternative approach could be to store the bundles which wrap all the referenced resources.


config/default.json
```diff
@@ -0,0 +1,3 @@
+{
+    "knowledge_artifact_servers": ["http://pathways.mitre.org:8190/fhir"]
```
dehall, line 2: Can we use the `servers` collection in the data store to store the list of servers instead? Then we could also store additional metadata like "time last fetched", credentials, or anything else we might find useful. Then maybe a separate task would be a command line function to register new servers

src/storage/impl/Internal.js
```diff
@@ -60,6 +60,15 @@ class Internal {
     collection.removeWhere(whereFn);
     // note that this does not return anything. we may need to update that at some point
   }
+
+  upsert(collectionName, value, whereFn) {
+    const resultList = this.select(collectionName, whereFn);
+    if (resultList.length > 0) {
+      this.update(collectionName, whereFn, () => value);
```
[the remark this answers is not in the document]
blangley28: Good call. Updated this and updated the put method in the servers route to use this method. Tested that it does indeed update now

src/storage/impl/Internal.js
```diff
@@ -24,6 +27,9 @@ class Internal {
         autosave: true,
         serializationMethod: 'pretty'
       });
+      if (fs.existsSync('medmorph.db')) {
```
dehall, line 30: The recommended approach in LokiJS to perform things immediately after initialization is to use the `autoloadCallback` option: https://github.com/techfort/LokiJS/wiki/LokiJS-persistence-and-adapters
I think the reason is the init process does some async stuff and may not be ready immediately after the constructor returns

src/routes/servers.js
```diff
@@ -52,11 +52,7 @@ router.put('/:id', (req, res) => {
   // for a more rigorous API we should either check first,
   // or use an "upsert"-type function within the DB
```
dehall, line 53: now that we are actually using upsert we can delete this comment

src/storage/DataAccess.js
```diff
@@ -32,6 +32,13 @@
  *   @param {string} collectionName - name of collection to delete from
  *   @param {function(object): boolean} whereFn - function to identify rows to delete
  *   @return {void}
+ *
+ * Insert or Update:
+ * - upsert(collectionName, value, whereFn)
+ *   @param {string} collectionName - name of the collection to add to
+ *   @param {object|array} value - object to add, or array of objects to add
```
dehall, line 39: I think we probably want this to just be `object` - not sure how the update operation would work given an array

package.json
```diff
@@ -21,13 +21,17 @@
     "supertest": "^6.0.1"
   },
   "dependencies": {
+    "axios": "^0.21.1",
+    "babel-eslint": "^10.1.0",
+    "config": "^3.3.3",
```
dehall, line 26: Seems like `config` isn't needed anymore? But on the other hand, just to confirm - babel-eslint is still needed even with the upgrade to node 14?

src/storage/impl/Internal.js
blangley28: I tried that last night and it didn't work. To make this work I had to pass in `this` instead of requiring it. With a fresh set of eyes on it this morning I realized that's what I needed to make the callback work

src/storage/DataAccess.js
blangley28: Agreed. I had copied this from the insert comments without even thinking about it

package.json
blangley28: Removed config but babel-eslint is still required to make lint pass

src/utils/fhir.js
```diff
@@ -14,4 +22,40 @@ function generateOperationOutcome(code, msg) {
...
+}
+
+/**
+ * Get all knowledge artifacts (from servers registered in the config
+ * file) and save them. Stores all refrenced resources as well.
+ */
+function refreshKnowledgeArtifacts(db) {
+  const servers = db.select('servers', s => s.type === 'KA');
+  servers.forEach(server => {
+    getResources(server.endpoint, 'PlanDefinition').then(data => {
```
dehall, line 50: One last super minor thing, as you address the lint issues. Can you add a debug log line here that reports the server it's fetching from?

dehall (approved):

dubem7 (approved):
Looks good to me

PR #6, opened 2021-01-21
Backend service client auths to servers

dubem7:
The app can make calls to EHR or KA using backend authentication.

We will need to remove indicated lines before merging, these were left to make reviewing the changes easier.


blangley28:
This should fix your lint error (same fix I had to do in my other branches). You need to add `babel-eslint` as a parser to the `.eslintrc.json` file.
1. `npm install babel-eslint`
2. Add this line right before `parserOptions` in `.eslintrc.json` 
```
"parser": "babel-eslint",
```

src/utils/client.js
```diff
@@ -0,0 +1,123 @@
...
+  constructor(jwks) {
+    this.jwks = jwks;
+  }
+
+  /**
+   * Generate and return access token for the specified server
+   * @param {*} server  the server to get an access token for
+   */
+  async connectToServer(server) {
+    if (server === 'EHR') {
```
dehall, line 24: Let's make this function take in a URL directly so that we don't have to maintain a separate list to look items up in

src/utils/privateKey.json
```diff
@@ -0,0 +1,18 @@
+{
```
dehall, line 1: Does anything break if extra info is added to this file? I think we should somehow more clearly indicate that this is a key generated just for this reference implementation, and it's not a leaked secret. If it doesn't break let's add a "_notes" or some kind of field at the top explaining that

src/utils/key.json
```diff
@@ -0,0 +1,11 @@
+{
```
dehall, line 1: This file isn't used anywhere in the code right? It just needed to be accessible to register the client? I'm a little worried we might see this isn't referenced then delete it. Can you make a new folder `src/keys` and have them be named "publicKey" and "privateKey" so the link is obvious?

src/utils/serverConfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
blangley28, line 1: I updated my PR to remove the config for the KA and just load it into the servers collection. We should do the same here and remove this config

src/utils/client.js
dubem7: This lets people put in whatever they want as opposed to a limited set that we known will work, is this an issue?

dehall: That's fine. In practice the connections are going to be made only to selected "known good" servers anyway, it's not like we're going to have a UI that users type in free text into

src/utils/privateKey.json
dubem7: I added a readme, I got an error after adding a "_notes" files

src/utils/key.json
dubem7: moved and renamed the files

src/utils/serverConfig.json
dubem7: Removed the file

src/utils/client.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Generate a signed JWT used for authenticating
+   * @param {*} client_id The identifier of the client on the remote server
+   * @param {*} aud The token url of the server the JWT is being created for
+   * @param {*} kid The identifier of the key in the JWKS to sign the JWT
+   */
+  async generateJWT(client_id, aud, kid = this.signingKeyId) {
+    const options = { alg: 'RS384', compact: true };
```
blangley28, line 90: The smart backend authorization spec allows for either RS384 or EC384 to be used. I'm fine with hard coding this to be RS384 but I think we should add a note and/or a new task

src/utils/client.js
```diff
@@ -0,0 +1,107 @@
...
+      }
+    };
+
+    let response = await axios.get(`${url}/.well-known/smart-configuration`);
+    const wellKnown = response.data.token_endpoint;
+    const jwt = await this.generateJWT(clientId, wellKnown, keys.kid);
+    props.client_assertion = jwt;
+    response = await axios.post(wellKnown, queryString.stringify(props), headers);
+    const accessToken = response.data.access_token;
+    headers.headers.Authorization = `Bearer ${accessToken}`;
```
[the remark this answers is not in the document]
dubem7: I believe so yes

src/utils/client.js
dubem7: Added a note and created a ticket for it

dehall (approved):
Looks good to me. After the rebase, please make sure to push any additional changes to package-lock.json at the end. When I ran `npm install` I had changes in package-lock

src/routes/index.js
```diff
@@ -15,3 +15,10 @@ router.post('/fetch-ka', (req, res) => {
 });
 
 module.exports = router;
+
+const client = require('../utils/client');
+
+const newClient = new client.Client(keys);
+
+newClient.connectToServer('http://pathways.mitre.org:8180/fhir');
+newClient.connectToServer('http://pathways.mitre.org:8190/fhir');
```
blangley28, line 24: These two lines are just for testing right? They should be removed before merging

src/utils/client.js
```diff
@@ -0,0 +1,82 @@
...
+  constructor(jwks) {
+    this.jwks = jwks;
+  }
+
+  /**
+   * Generate and return access token for the specified server
+   * @param {string} url  the fhir base url for the server to connect to
+   */
+  async connectToServer(url) {
+    let response = await axios.get(`${url}/.well-known/smart-configuration`);
```
blangley28, line 21: This gives a lint warning about using const instead of let
```suggestion
    const response = await axios.get(`${url}/.well-known/smart-configuration`);
```

blangley28 (approved):

PR #7, opened 2021-02-01
Turn KnowldegeArtifacts into Subscriptions

blangley28:
Util functions to convert a KA Bundle or a single PlanDefinition into the necessary Subscriptions. This uses the R5 Backport for Subscriptions. Since `SubscriptionTopic`s don't exist in R4 the topic is defined via the canonical url and then it is up to the implementing server to know what those map to.

### DIFF
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -164,15 +164,27 @@ function subscriptionsFromPlanDef(planDef, url, token) {
 
   if (action.trigger?.length === 0) return [];
 
-  return action.trigger.map(trigger => {
+  return action.trigger.reduce((subscriptions, trigger) => {
     const namedEventExt = trigger.extension.find(e => e.url === NAMED_EVENT_EXTENSION);
     const namedEventCoding = namedEventExt.valueCodeableConcept.coding.find(
       c => c.system === NAMED_EVENT_CODE_SYSTEM
     );
 
     const criteria = namedEventToCriteria(namedEventCoding.code);
-    if (criteria) return generateSubscription(namedEventCoding.code, criteria, url, token);
-  });
+    if (criteria)
+      subscriptions.push(generateSubscription(namedEventCoding.code, criteria, url, token));
+    return subscriptions;
+  }, []);
+
+  // return action.trigger.map(trigger => {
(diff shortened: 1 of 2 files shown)


src/utils/fhir.js
```diff
@@ -60,4 +134,92 @@ function refreshKnowledgeArtifacts(db) {
...
+  if (action.trigger?.length === 0) return [];
+
+  return action.trigger.map(trigger => {
+    const namedEventExt = trigger.extension.find(e => e.url === NAMED_EVENT_EXTENSION);
+    const namedEventCoding = namedEventExt.valueCodeableConcept.coding.find(
+      c => c.system === NAMED_EVENT_CODE_SYSTEM
+    );
+
+    const criteria = namedEventToCriteria(namedEventCoding.code);
+    if (criteria) return generateSubscription(namedEventCoding.code, criteria, url, token);
```
dehall, line 174: If `criteria` is null then this inner function returns undefined, and we'll probably have to check for that in the array downstream. Can this be reworked to ensure the function doesn't pull null/undefined into the list? Or another quick approach is just tack `.filter(x => x)` after the `.map(...)` call to filter out any falsy values.

The other question this raises for me is, should there be some way of indicating that we have a PlanDef that we can't create a subscription for, because we can't map it to criteria here? Maybe for now just write to debug output that it happened?

blangley28: Updated the map to be a reduce and wrote a new test to verify it works

src/utils/fhir.js
```diff
@@ -60,4 +134,104 @@ function refreshKnowledgeArtifacts(db) {
...
+      c => c.system === NAMED_EVENT_CODE_SYSTEM
+    );
+
+    const criteria = namedEventToCriteria(namedEventCoding.code);
+    if (criteria)
+      subscriptions.push(generateSubscription(namedEventCoding.code, criteria, url, token));
+    return subscriptions;
+  }, []);
+
+  // return action.trigger.map(trigger => {
```
dehall, line 179: Some leftover commented code here

blangley28: Whoops. Removed

dehall (approved):

dubem7 (approved):

PR #8, opened 2021-02-01
action implementation

KeeyanGhoreshi:
This is a WIP implementation of the actions listed here:

https://build.fhir.org/ig/HL7/fhir-medmorph/ValueSet-us-ph-plandefinition-action.html

There are quite a few things that need some work or are up in the air about implementation.  Lots of comments throughout the code as well.  

One thing I'll note here is that I realized `extract-research-data` could be filtered down through the FHIR query, rather than filtered down through the client code, a bit late and implemented it strangely.  I will likely go back and switch it up to make a fhirpath query, granted that's even possible.  

Other things that aren't exactly clear:

`check-trigger-codes`: what exactly is being triggered?  

`checkParticipantRegistration`: the planDef can define multiple conditions (per action!), and there are multiple participants. If this is really supposed to be `checkReportability`, then will the same thing apply but without the check between participants?

`validateReport`: each action supports a profile parameter, so we could potentially validate the report based on that.  There are multiple outputs allowed, each with multiple profiles allowed, so are we validating the entire bundle across all the profiles?  We could do it on a resourceType basis by making multiple outputs and binding them each to a different resourceType.  This is how I tentatively implemented it, by cycling through all the outputs and checking the profile.

Technically profile can be an array, but I'm not sure how that would work.

`encryptReport`: I wasn't sure what we wanted for encryption.  Base64 encoding it and tossing it into a binary resource is simple enough.  Should we introduce a library for some type of encryption?

`extractResearchData`:  Like I mentioned, I'll probably go back and switch it to make a more complicated FHIR query, rather than do the filtering ourselves inside the client.

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -14,8 +14,6 @@
 //      database
 //  reportingBundle: the entire bundle to be submitted
 //  contentBundle: just the content section of the report bundle
-//  next: place to put arbitrary objects specifically for the next step?
-//  prev: where arbitrary objects from the previous set should go?
 //  flags: various boolean or enum flags about the status of the report.
 //         is it encrypted y/n, is it identifiable or de/pseudo/anonymized, etc
 //  cancelToken: indicates the job should be cancelled (from a process external
@@ -61,14 +59,14 @@ const baseIgActions = {
   },
   'create-report': context => {
     // the output might define a profile
-    let contentBundle = createBundle(context.records);
+    let contentBundle = createBundle(context.records, 'collection');
     if (context.action.output) {
       const output = context.action.output[0];
       if (output.profile) contentBundle = validateProfile(contentBundle, output.profile[0]);
     }
     const header = makeHeader(context);
     context.contentBundle = contentBundle;
-    context.reportingBundle = createBundle([header, ...contentBundle.entry]);
+    context.reportingBundle = createBundle([header, ...contentBundle.entry], 'message');
   },
   'validate-report': context => {
     // check if fhir is well-formed
@@ -80,8 +78,7 @@ const baseIgActions = {
       .submit(context.reportingBundle)
       .then(
         result => {
-          // might want to check for a 200?
-          if (result.status === 200) {
+          if (result.status === 200 || result.status === 202) {
             context.flags['submitted'] = true;
           }
         },
@@ -99,7 +96,7 @@ const baseIgActions = {
       .deidentify(context.reportingBundle)
       .then(
         result => {
-          context.reportingBundle = result;
+          context.reportingBundle = result.data;
           context.flags['deidentified'] = true;
         },
         () => {
@@ -116,7 +113,7 @@ const baseIgActions = {
       .anonymize(context.reportingBundle)
       .then(
         result => {
-          context.reportingBundle = result;
+          context.reportingBundle = result.data;
           context.flags['anonymized'] = true;
         },
         () => {
@@ -133,7 +130,7 @@ const baseIgActions = {
       .pseudonymize(context.reportingBundle)
       .then(
         result => {
-          context.reportingBundle = result;
+          context.reportingBundle = result.data;
           context.flags['pseudonymized'] = true;
         },
         () => {
@@ -251,17 +248,16 @@ function getBySearchParam(resource, searchParam) {
 }
 
 function compareDates(date, period) {
-  // TODO: returns true if the periods overlap
   const now = Date.parse(date);
   const start = Date.parse(period.start);
   const end = Date.parse(period.end);
   return now > start && now < end;
 }
 
-function createBundle(records) {
+function createBundle(records, type) {
   const bundle = {
     resourceType: 'Bundle',
-    type: 'message',
+    type: type,
     timestamp: Date.now(),
     entry: []
   };
@@ -303,7 +299,7 @@ function makeHeader(context) {
       }
     ],
     source: {
-      endpoint: context.client.source.endpoint
+      endpoint: `${process.env.BASE_URL}/$process-message`
     },
     sender: {
       // would the org come from the plandef?
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -21,355 +21,376 @@
 //  cancelToken: indicates the job should be cancelled (from a process external to the job itself -- this tells the job to cease processing)
 //  exitStatus: if set, terminates the job and marks it with the given status (for example, if a given patient doesn't meet reporting criteria, don't keep reporting)
 
-
-// we assume the context updates and is passed to functions 
+// we assume the context updates and is passed to functions
 // in turn.  I'm also assuming the context is unique
-// to the current action, since we'd be cycling through actions 
+// to the current action, since we'd be cycling through actions
 // anyway.
 
 function checkTriggerCodes(context) {
-    // are we checking the action or cycling through all actions in the plandef?
-    const action = context.action;
-    const triggers = action.trigger;
-    const triggerCodes = []
-    triggers.map((trigger) => {
-        if (trigger.type === 'named-value' && trigger.extension) {
-            const triggerEvents = trigger.extension.filter((ext) => {
-                return ext.url === 'http://hl7.org/fhir/us/medmorph/StructureDefinition/ext-us-ph-namedEventType';
-            });
+  // are we checking the action or cycling through all actions in the plandef?
+  const action = context.action;
+  const triggers = action.trigger;
+  const triggerCodes = [];
+  triggers.map(trigger => {
+    if (trigger.type === 'named-value' && trigger.extension) {
+      const triggerEvents = trigger.extension.filter(ext => {
+        return (
+          ext.url === 'http://hl7.org/fhir/us/medmorph/StructureDefinition/ext-us-ph-namedEventType'
+        );
+      });
 
-            // TODO: What would the correct way to get 
-            // the code from the extension be?
-            const triggerCode = triggerEvents.valueCode;
-            if (triggerCode === 'encounter-change' || triggerCode === 'encounter-modified') {
-                // these are basically the same, but encounter-change is lvl 0
-                // not sure what that means for this
-                triggerCodes.push(triggerCode);  // no need to check the encounter here
-            } else if (triggerCode === 'encounter-start') {
-                if (context.encounter.period && context.encounter.period.low) {
-                    // TODO: check if it's past the start time?  
-                    triggerCodes.push(triggerCode);
-                }
-            } else if (triggerCode === 'encounter-stop') {
-                if (context.encounter.period && context.encounter.period.high) {
-                    // TODO: check if it's past the end time?  
-                    triggerCodes.push(triggerCode);
-                }
-            } else if (triggerCode === 'diagnosis-change' 
-            || triggerCode === 'new-diagnosis' 
-            || triggerCode === 'modified-diagnosis') {
-                // not sure about this one, just check for a resource
-                const conditions = context.records.filter((resource) => {return resource.resourceType === 'Condition'});
-                if (conditions.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            } else if (triggerCode === 'medication-change' 
-            || triggerCode === 'new-medication' 
-            || triggerCode === 'modified-medication') {
-                // not sure about this one, just check for a resource
-                const medications = context.records.filter((resource) => {
-                    return resource.resourceType === 'MedicationRequest' 
-                    || resource.resourceType === 'MedicationDispense'
-                    || resource.resourceType === 'MedicationStatement'
-                    || resource.resourceType === 'MedicationAdministration'
-                });
-                if (medications.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            }  else if (triggerCode === 'labresult-change' 
-            || triggerCode === 'new-labresult' 
-            || triggerCode === 'modified-labresult') {
-                // not sure about this one, just check for a resource
-                const observations = context.records.filter((resource) => {return resource.resourceType === 'Observation'});
-                if (observations.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            }  else if (triggerCode === 'order-change' 
-            || triggerCode === 'new-order' 
-            || triggerCode === 'modified-order') {
-                // not sure about this one, just check for a resource
-                const orders = context.records.filter((resource) => {return resource.resourceType === 'ServiceRequest'});
-                if (orders.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            }  else if (triggerCode === 'procedure-change' 
-            || triggerCode === 'new-procedure' 
-            || triggerCode === 'modified-procedure') {
-                // not sure about this one, just check for a resource
-                const procedures = context.records.filter((resource) => {return resource.resourceType === 'Procedure'});
-                if (procedures.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            }  else if (triggerCode === 'immunization-change' 
-            || triggerCode === 'new-immunization' 
-            || triggerCode === 'modified-immunization') {
-                // not sure about this one, just check for a resource
-                const immunizations = context.records.filter((resource) => {return resource.resourceType === 'Immunization'});
-                if (immunizations.length > 0) {
-                    triggerCodes.push(triggerCode);
-                }
-            } else if (triggerCode === 'demographic-change') {
-                // just make sure the patient has demographics?
-                const patient = context.patient;
-                if (patient.gender 
-                    || patient.birthDate 
-                    || patient.name 
-                    || patient.deceasedBoolean 
-                    || patient.address 
-                    || patient.telecom) {
-                        triggerCodes.push(triggerCode);
-                    }
-            } else if (triggerCode === 'received-public-health-report' 
-            || triggerCode === 'received-public-health-response' 
-            || triggerCode === 'manual-notification') {
-                // no idea about these
-                triggerCodes.push(triggerCode);
-            }
+      // TODO: What would the correct way to get
+      // the code from the extension be?
+      const triggerCode = triggerEvents.valueCode;
+      if (triggerCode === 'encounter-change' || triggerCode === 'encounter-modified') {
+        // these are basically the same, but encounter-change is lvl 0
+        // not sure what that means for this
+        triggerCodes.push(triggerCode); // no need to check the encounter here
+      } else if (triggerCode === 'encounter-start') {
+        if (context.encounter.period && context.encounter.period.low) {
+          // TODO: check if it's past the start time?
+          triggerCodes.push(triggerCode);
         }
-    });
-    // assuming the flags are an object with keys
-    // for each flag, and that we can put in
-    // an array of fixed values.
-    context.flags['triggerCodes'] = triggerCodes;
-    return context;
+      } else if (triggerCode === 'encounter-stop') {
+        if (context.encounter.period && context.encounter.period.high) {
+          // TODO: check if it's past the end time?
+          triggerCodes.push(triggerCode);
+        }
+      } else if (
+        triggerCode === 'diagnosis-change' ||
+        triggerCode === 'new-diagnosis' ||
+        triggerCode === 'modified-diagnosis'
+      ) {
+        // not sure about this one, just check for a resource
+        const conditions = context.records.filter(resource => {
+          return resource.resourceType === 'Condition';
+        });
+        if (conditions.length > 0) {
+          triggerCodes.push(triggerCode);
+        }
+      } else if (
+        triggerCode === 'medication-change' ||
+        triggerCode === 'new-medication' ||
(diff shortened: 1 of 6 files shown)


src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+
+// we assume the context updates and is passed to functions
+// in turn.  I'm also assuming the context is unique
+// to the current action, since we'd be cycling through actions
+// anyway.
+
+function checkTriggerCodes(context) {
+  // are we checking the action or cycling through all actions in the plandef?
+  const action = context.action;
+  const triggers = action.trigger;
```
blangley28, line 32: So Dylan and I were working off the assumption there would be only be a trigger on the first action (note this isn't enforced by the profile). If you take a look at the [Cancer Reporting Example](https://build.fhir.org/ig/HL7/fhir-medmorph/PlanDefinition-plandefinition-cancer-example.json.html) you can see the first action has the triggers but the next action is for check-trigger-code. With this in mind it seems like perhaps the context should be initialized with the triggers (a nice side effect is that this list on the context could be just the codes instead of the entire trigger object)

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+  const triggerCodes = [];
+  triggers.map(trigger => {
+    if (trigger.type === 'named-value' && trigger.extension) {
```
blangley28, line 35: This is a fixed required value on the profile so I don't think we need to check here. If we're worried about the KAs not being conformant we should do that check when we fetch them not use them

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+      const triggerEvents = trigger.extension.filter(ext => {
+        return (
+          ext.url === 'http://hl7.org/fhir/us/medmorph/StructureDefinition/ext-us-ph-namedEventType'
```
blangley28, line 38: Just a thought - I used this same extension URL in my PR and I define it at a constant. Do you think it would be worth creating a new module with all the constant urls etc which the different files can then reference?

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+        );
+      });
+
+      // TODO: What would the correct way to get
+      // the code from the extension be?
+      const triggerCode = triggerEvents.valueCode;
```
blangley28, line 44: Looking at the trigger extension I think it would be  
```suggestion
      const triggerCode = triggerEvents.valueCodeableConcept.coding[0].code;
```

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+        return true;
+      }
+    });
+  }
+}
+
+function createReport(context) {
+  context.contentBundle = createBundle(context.records);
+  const reportingBundle = createBundle(context.records);
+  // assume the context has a messageHeader.
```
blangley28, line 197: What is this messageHeader? (where did it come from and what does it say?)

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+
+  if (isValid) {
+    context.flags['valid'] = true;
+  } else {
+    context.flags['valid'] = false;
+  }
+}
+
+function submitReport(context) {
+  context.client.dest.submit(context.reportingBundle);
```
blangley28, line 227: Where are all of these client functions going to be defined?

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+}
+
+function deidentifyReport(context) {
+  const client = context.client.trustServices;
+  client.deidentify(context.contentBundle);
```
blangley28, line 232: All of these trust service functions return a bundle. Since they are all http calls to another server they're going to need to be async to get the response bundle. That response should then be saved to the context as the bundle

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+
+function pseudonymizeReport(context) {
+  const client = context.client.trustServices;
+  client.pseudonymize(context.contentBundle);
+}
+
+function encryptReport(context) {
+  // TODO: Are we meant to actually encrypt the bundle?
+  // do we need an external library to encrypt the
+  // stringified bundle?
```
blangley28, line 248: I feel like we can just make this a `nop` because using TLS will automatically encrypt it for us. I don't know how else we would encrypt the bundle such that the receiving party could decrypt it (what other shared secret or asymmetric keys exist to encrypt? I don't think any do)

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+  const bundle = context.reportingBundle;
+  const base64 = btoa(JSON.stringify(bundle));
+  const resource = {
+    resourceType: 'Binary',
+    type: 'application/json',
+    data: base64
+  };
+  context.encryptedBundle = resource;
```
blangley28, line 256: The way I envisioned this is a flow from one of these actions to another. Each one is going to manipulate the report until its submitted and the workflow is done. As such I don't think these should all be saved as different types of bundles in the context. There's no point encrypting it and saving it to `context.encryptedBundle` if no other action uses that (and since an action will not know which other actions have been called previously you can't guarantee this property has been set). So I think all of these should be saving the updated bundle to the same property of context.

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+  // TODO: what's a datamart?
+}
+function getByPath(resource, path) {
+  // TODO: function to search resources by path
+  // and return correct attribute
+  return resource;
+}
+
+function compareCodes(code1, code2) {
+  // assume they're both coding
```
blangley28, line 350: Might as well name it as such
```suggestion
function compareCodes(coding1, coding2) {
```

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+    bundle.push({
+      resource: record
+    });
+  });
+
+  return bundle;
+}
+
+function getPatient(context) {
+  // assume the patient is only an ID
```
blangley28, line 380: I don't think this is an assumption. Based on the context model its a fact

src/utils/actions.js
```diff
@@ -0,0 +1,396 @@
...
+  const patientId = context.patient;
+  const patient = context.client.source.read(patientId);
+  context.patient = patient;
+  return context;
+}
+
+function evaluateExpression(expression, resource) {
```
blangley28, line 387: Perhaps I'm making this up but I thought I remember Rob saying MedMorph was trying to go fully fhir path and include no cql

blangley28 (commented):
I think you brought up lots of great questions and many of them are stumping me too. Would you be able to compile all of them into a list? I'm not sure if any of us will have answers to your questions so we will need to ask Dragon on Zulip (although he's yet to respond to any of Rob's questions) or bring it up on the Wednesday WG meeting.

Also, it looks like you updated the context a little bit. Not sure if you want to do it now or before you merge but could you please update the wiki?

src/utils/actions.js
dehall: Yeah the latest was that the spec allows both but requires FHIRPath. So at that point there's really no reason to also include CQL. But again the spec does allow it

dehall:
Re: updating the wiki to reflect what's in `context`, let's do that once we have a relatively stable set of things we want to include

dehall:
One piece that is missing here is the "Extensibility" component.  https://github.com/mcode/medmorph-backend/wiki/Reporting-Workflow#extensibility

There are a couple ways to do this, and tbh we may wind up switching things around over time, but basically we need a way for content IGs to specify additional codes or override the implementation of existing codes.
I think the easiest way is to instead of having all these functions be at the top-level, have them be properties of an object, ex
```
function checkTriggerCodes(context) {
  ....
}
```
becomes -->
```
const baseIgActions = {
  "check-trigger-codes": context => { ... },
  
  "do-the-other-thing": ...
}
```

That way content IGs can "extend" that in one line, and if we need to make it more complicated like an ES6 class we can do that later.

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+    if (context.action.output) {
+      const output = context.action.output[0];
+      if (output.profile) {
+        profile = output.profile[0];
+      }
+    }
+    let contentBundle = createBundle(context.records);
+    if (profile) {
+      contentBundle = validateProfile(contentBundle, profile);
+    }
```
blangley28, line 69: Since `profile` isn't used anywhere else this can be condensed to 
```suggestion
    if (context.action.output) {
      const output = context.action.output[0];
      if (output.profile)
        contentBundle = validateProfile(contentBundle, output.profile[0]);
    }
    let contentBundle = createBundle(context.records);
```

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+  },
+  'submit-report': context => {
+    context.client.dest.submit(context.reportingBundle);
+  },
+  'deidentify-report': context => {
+    const client = context.client.trustServices;
+    client.deidentify(context.reportingBundle).then(result => {
+      context.reportingBundle = result;
+    });
+    context.flags['deidentified'] = true;
```
blangley28, line 87: I think these trustService methods should be async. The deidentified flag will be set to true before the `context.reportingBundle` is updated. If there's no delay between this and the next action then its highly likely to be overwritten. 
This should also be coordinated with Dylan because his code is expecting these actions to run synchronously.

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+      context.reportingBundle = result;
+    });
+    context.flags['pseudonymized'] = true;
+  },
+  'encrypt-report': context => {
+    // noop
+    context.flags['encrypted'] = true;
+  },
+  'complete-reporting': context => {
+    context.client.db.insert('completed reports', context.reportingBundle);
```
blangley28, line 108: Since this is a new collection which isn't a FHIR type could you please add it to the [Collections wiki](https://github.com/mcode/medmorph-backend/wiki/Collections)

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+      });
+    }
+  },
+  'execute-research-query': context => {
+    // noop
+  }
+};
+
+function validateProfile(report, structureDefinition) {
+  // TODO: Perhaps this is something one of the servers/clients would handle?
```
blangley28, line 187: Is it not possible to use `fhir.validate` like you did above?

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+  return report;
+}
+
+function getByPath(resource, path) {
+  // TODO: function to search resources by path
```
blangley28, line 192: Is this...fhirPath?

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+function getBySearchParam(resource, searchParam) {
+  // TODO: function to search resources by param
+  // and return correct attribute
+  return resource.date;
+}
+
+function compareDates(period1, period2) {
+  // TODO: returns true if the periods overlap
+}
+function createBundle(records) {
```
blangley28, line 211: Can we make this conformant to the two bundle profiles (shouldn't be difficult based on the differential table)
[Public Health Reporting Bundle](https://build.fhir.org/ig/HL7/fhir-medmorph/StructureDefinition-us-ph-reporting-bundle.html)
[Public Health Content Bundle](https://build.fhir.org/ig/HL7/fhir-medmorph/StructureDefinition-us-ph-content-bundle.html)

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+  });
+
+  return bundle;
+}
+
+function getPatient(context) {
+  const patientId = context.patient;
+  const patient = context.client.source.read(patientId);
+  context.patient = patient;
+  return context;
```
blangley28, line 231: Is there a reason `context.patient` is just the identifier but `context.encounter` is the Encounter? Unless there's a good reason I propose changing to make the `context.patient` the entire Patient resource. It is then the responsibility of the subscription notification handler to retrieve the patient while initializing the context.

If it should remain as just the identifier, wouldn't the patient exist in `context.records`? So it could be found from there instead of making a network call.

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+
+function evaluateExpression(expression, resource) {
+  if (expression.language === 'text/fhirpath') {
+    const fhirpath = new FhirPath(resource);
+    const path = fhirpath.evaluate(expression.expression);
+    return path;
+  }
+}
+
+function makeHeader(profile) {
```
blangley28, line 242: This is missing some important elements of the [Public Health MessageHeader](https://build.fhir.org/ig/HL7/fhir-medmorph/StructureDefinition-us-ph-messageheader.html), such as source and destination.

blangley28 (commented):
Have you tested any of these? I know this task said assume the context is given but would it be possible to write some tests for all of these functions using a mock context (especially for context.client).

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+// anyway.
+const Fhir = require('fhir').Fhir;
+const FhirPath = require('fhir').FhirPath;
+const fhir = new Fhir();
+const baseIgActions = {
+  'check-trigger-codes': context => {
+    const action = context.action;
+    const resources = context.records;
+
+    const boolMap = action.condition.map(condition => {
```
blangley28, line 35: Looking at the [example](https://build.fhir.org/ig/HL7/fhir-medmorph/PlanDefinition-plandefinition-cancer-example.json), I think this should be `action.action.condition`.

src/utils/actions.js
dehall: The whole reporting workflow can be made async without too much trouble. I just made a quick change to mine that assumes each action can be async so it's called with await.

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+      if (condition.kind === 'applicability' && condition.expression) {
```
blangley28, line 36: Something else I just noticed too is the condition.kind always needs to be applicability https://build.fhir.org/ig/HL7/fhir-medmorph/provisioning.html#creating-knowledge-artifacts

src/utils/actions.js
dehall: Yeah we should be consistent if possible. Either both should be identifiers or both should be resources, whichever makes more sense

src/utils/actions.js
```diff
@@ -0,0 +1,254 @@
...
+//      dest: FHIR client to send to PHA/TTP
+//      trust services
+//      database
+//  reportingBundle: the entire bundle to be submitted
+//  contentBundle: just the content section of the report bundle
+//  next: place to put arbitrary objects specifically for the next step?
+//  prev: where arbitrary objects from the previous set should go?
+//  flags: various boolean or enum flags about the status of the report. is it encrypted y/n, is it identifiable or de/pseudo/anonymized, etc
+//  cancelToken: indicates the job should be cancelled (from a process external to the job itself -- this tells the job to cease processing)
+//  exitStatus: if set, terminates the job and marks it with the given status (for example, if a given patient doesn't meet reporting criteria, don't keep reporting)
```
blangley28, line 21: Based on the actions here can messageHeader, next, and prev be removed? Content IGs may want next and prev

src/utils/actions.js
```diff
@@ -0,0 +1,325 @@
...
+    const validation = fhir.validate(context.reportingBundle);
+    context.flags['valid'] = validation.valid;
+  },
+  'submit-report': async context => {
+    await context.client.dest
+      .submit(context.reportingBundle)
+      .then(
+        result => {
+          // might want to check for a 200?
+          if (result.status === 200) {
```
blangley28, line 84: So our implementation definitely returns 200 on success but the $process-message docs allow for 202 Accepted to be returned. Since we're not using the response data here is it worth checking for 202 as well?

src/utils/actions.js
```diff
@@ -0,0 +1,325 @@
...
+//  Patient: identifier for the patient whose data triggered the workflow
+//  records: list of records
+//  Encounter: encounter where some data triggered the workflow
+//  Organization: Organization that is doing the workflow
+//  PlanDef: *** do we want to keep a "pre-processed" version of the plandef?
+//  Action: action currently in context from the planDef
+//  client: set of clients to connect to
+//      source: FHIR client to make additional queries to EHR
+//      dest: FHIR client to send to PHA/TTP
+//      trust services
```
blangley28, line 13: We should sync up on these. It looks like your `.submit` is assuming the entire axios object (with the `result.status` ) property, but the data trust service requests are expecting just the data which comes from `result.data`). 
My current implementation of these functions is to just return the data, but since you use other parts and handle errors here I'm leaning towards all of these networking calls to just return the axios promise. That means for this PR `result` needs to be updated to `result.data` on
- [ ] line 102
- [ ] line 119
- [ ] line 136

src/utils/actions.js
```diff
@@ -0,0 +1,325 @@
...
+      }
+    ],
+    eventCoding: {
+      system: 'http://hl7.org/fhir/us/medmorph/CodeSystem/us-ph-messageheader-message-types',
+      code: 'message-report'
+    },
+    destination: [
+      {
+        // assume client has some way to produce its endpoint
+        endpoint: context.client.dest.endpoint
```
blangley28, line 302: I can add this to the context since I am already pulling it out for the `context.client.dest.submit` operation. Lets make sure I collect all your changes. I requested your review on my PR

src/utils/actions.js
```diff
@@ -0,0 +1,325 @@
...
+      }
+    ],
+    source: {
+      endpoint: context.client.source.endpoint
```
blangley28, line 306: These two sources are not the same. The `MessageHeader.source` is the source of the message (aka this application). The `context.client.source` is the data source for this app (aka the EHR). We should double check but I'm pretty sure this needs to be the $process-message endpoint of this app so it can receive responses back

src/utils/actions.js
blangley28: It seems like the only change to be made here is the type needs to be `collection` for the content bundle

src/utils/actions.js
blangley28: Looks like `context.patient` isn't even used anymore in the latest commit. Based on the tests though it has been updated to be the complete patient instead of just the id

src/utils/actions.js
KeeyanGhoreshi: the validate function doesn't support custom profile validation, unfortunately, as I've recently found out.  

See: https://github.com/lantanagroup/FHIR.js/issues/29

src/utils/actions.js
KeeyanGhoreshi: The patient will probably be needed for `check-participant-registration` but I'm still not quite sure what that actions implementation is meant to be

src/utils/actions.js
KeeyanGhoreshi: the `testContext` resource has all the object elements needed for the actions, so that should be a good guide.

blangley28 (approved):

src/utils/actions.js
```diff
@@ -0,0 +1,321 @@
...
+  'create-report': context => {
+    // the output might define a profile
+    let contentBundle = createBundle(context.records, 'collection');
+    if (context.action.output) {
+      const output = context.action.output[0];
+      if (output.profile) contentBundle = validateProfile(contentBundle, output.profile[0]);
+    }
+    const header = makeHeader(context);
+    context.contentBundle = contentBundle;
+    context.reportingBundle = createBundle([header, ...contentBundle.entry], 'message');
```
dehall, line 69: The reporting bundle is supposed to contain the actual content bundle as a Bundle. So a Bundle inside a Bundle: https://build.fhir.org/ig/HL7/fhir-medmorph/StructureDefinition-us-ph-reporting-bundle.html#introduction

Also both of these bundles should have a profile on them, maybe that should be another arg to the create function

```suggestion
    context.reportingBundle = createBundle([header, contentBundle], 'message');
```

src/utils/actions.js
```diff
@@ -0,0 +1,321 @@
...
+    // check if fhir is well-formed
+    const validation = fhir.validate(context.reportingBundle);
+    context.flags['valid'] = validation.valid;
+  },
+  'submit-report': async context => {
+    await context.client.dest
+      .submit(context.reportingBundle)
+      .then(
+        result => {
+          if (result.status === 200 || result.status === 202) {
```
dehall, line 81: We do already have the `http-status-codes` dependency installed so that could be used here. See src/routes/servers.js for example usage

src/utils/actions.js
```diff
@@ -0,0 +1,321 @@
...
+    const client = context.client.trust;
+    client
+      .pseudonymize(context.reportingBundle)
+      .then(
+        result => {
+          context.reportingBundle = result.data;
+          context.flags['pseudonymized'] = true;
+        },
+        () => {
+          context.flags['pseudonymize'] = false;
```
dehall, line 137: This key doesn't match up with the one right above it. Same with the pseudonymize a few lines below.

src/utils/actions.js
blangley28: I think you can tell based on the type right? Or perhaps instead of passing in 'message' vs 'collection' the third parameter is 'reporting' vs 'content' and that can be used to set the Bundle.type and the profile

dehall (approved):

PR #10, opened 2021-02-04
Authorization

blangley28:
Implement SMART Backend Authorization Protocol. Requests to the backend app are now protected (using passport middleware) and require an access token from Keycloak. For testing purposes "admin" is also hard coded as a valid token like it is in the FHIR servers.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -24,8 +24,8 @@ app.use(passport.initialize());
 passport.use(
   new KeycloakBearerStrategy(
     {
-      realm: 'backend_app',
-      url: 'http://moonshot-dev.mitre.org:8090/auth'
+      realm: process.env.REALM,
+      url: process.env.AUTH
     },
     (jwtPayload, done) => {
       return done(null, jwtPayload);
--- a/src/app.js
+++ b/src/app.js
@@ -3,21 +3,43 @@ require('dotenv').config();
 const express = require('express');
 const logger = require('morgan');
 const bodyParser = require('body-parser');
+const passport = require('passport');
+const KeycloakBearerStrategy = require('passport-keycloak-bearer');
 
 const fhirRouter = require('./routes/fhir');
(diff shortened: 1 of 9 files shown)


src/utils/auth.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Helper function to obtain the token from the authorization header
+ *
+ * @param {*} req - the express request
+ * @returns the bearer token if it exists, otherwise null
+ */
+function getToken(req) {
+  const authHeader = req.headers.authorization;
+  if (authHeader) return authHeader.split(' ')[1];
```
dehall, line 11: Might be good just as a reminder to comment that this expects the header to be of the form "Bearer {token}".

src/app.js
```diff
@@ -3,21 +3,43 @@ require('dotenv').config();
...
 app.use(logger('dev'));
 app.use(express.json());
 app.use(bodyParser.json({ type: ['application/json', 'application/fhir+json'] }));
+app.use(passport.initialize());
+
+passport.use(
+  new KeycloakBearerStrategy(
+    {
+      realm: 'backend_app',
+      url: 'http://moonshot-dev.mitre.org:8090/auth'
```
dehall, line 28: Can this url and the token endpoint in `src/routes/wellknown.js` be made configurable? Those would work well in a .env file. Maybe we should also have a .env.dev or .env.mitre or something with these pre-populated.

blangley28: Added REALM and AUTH env vars which are required. Pushed .env.dev which contains an example .env file

dehall (approved):

KeeyanGhoreshi:
is there a way to test this without using the test `admin` token?

KeeyanGhoreshi (approved):
creating a JWT and posting to the auth server returns an auth token that the backend server accepts.  Works and the code looks fine.

PR #11, opened 2021-02-08
Persist authorization credentials

dubem7:
Persist data needed for authorization. 
Check that the token is still valid before returning the token.


src/utils/AuthorizationPersistance.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ */
+
+class AuthorizationPersistance {
+  addServer(server) {
+    db.insert(SERVERS, server);
+  }
+
+  addServerConfiguration(server, config) {
+    db.upsert(SERVERS, config, s => s.id === server.id);
```
blangley28, line 37: This will update the value in the database to just be `config`. But I think in reality it should update to be 
```
{
   ...server
   ...config
}
```

src/utils/AuthorizationPersistance.js
```diff
@@ -0,0 +1,89 @@
+const db = require('../storage/DataAccess');
+
+const SERVERS = 'serverConfig';
+const CLIENTS = 'clientConfig';
```
blangley28, line 4: These are two different collections right? I think we only need one (which is called `servers` right now). The app doesn't need to know the `jwks_url` of the clients registered with it. This is all handled via Keycloak. What the app does need to know is the servers it has registered as a client with. 

For example:
EHR registers with Backend App via Keycloak. The app itself does not need to know anything about this. The Backend App then registers with the EHR (shares its own jwks_url which is just `{backend app base url}/jwks`). The EHR will then return a client ID. The Backend app will then use this when creating the JWT client_assertion and obtain an access token. So the new data to store in the `servers` collection is `clientId` and `token`. 

Note there is already a [schema for the collection on the wiki](https://github.com/mcode/medmorph-backend/wiki/Collections) (certainly open to new ideas on now to present this since we're not using TypeScript). So the new schema would be

```js
{
   name: String // human readable name
   endpoint: String // FHIR base url
   type: 'KA' | 'EHR' | 'PHA' // type of server the endpoint is
   clientId: String // the client ID assigned to this app during registration
   token: String // the last access token received from authorization
   tokenExp: number // the time the access token expires
}
```

With all of that said my comments for change are:
- [ ] replace this with a more generic `servers` data access layer (i.e. `getServerClientId(serverId)`)
- [ ] link this to your `Client` class to get the `clientId` for `generateJWT` method
- [ ] update `connectToServer` in `Client` class to read/write the `token` and `tokenExp`

Note: since these recommendations are substantial changes I believe we agreed to get a second opinion before implementing them.

src/routes/servers.js
```diff
@@ -3,10 +3,12 @@ const { StatusCodes } = require('http-status-codes');
 
 const express = require('express');
 const router = express.Router();
+const servers = require('../storage/servers');
 
 const db = require('../storage/DataAccess');
 
 const COLLECTION = 'servers';
```
blangley28, line 10: Wont let me add a comment on line 16 so I'll do it here... might as well replace the get all servers with a method in the other servers.js file. Perhaps rename `getServer(serverId)` to `getServerById(serverId)` and then implement `getServers()` to return all.

src/storage/servers.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+
+class Servers {
+  addServer(server) {
+    db.upsert(SERVERS, server, s => s.id === server.id);
+  }
+
+  getServer(id) {
+    return db.select(SERVERS, s => s.id === id)[0];
```
blangley28, line 27: Will this cause issues if the server with the id doesn't exit? I'm assuming it returns an empty list so `[][0]` will be undefined but just want to double check. Perhaps add a test?

src/storage/servers.js
```diff
@@ -0,0 +1,85 @@
...
+ *  @param {string} endpoint: - FHIR base url
+ *  @param {string} type - 'KA' | 'EHR' | 'PHA' - type of server the endpoint is
+ *  @param {string} clientId - the client ID assigned to this app during registration
+ *  @param {string} token - the last access token received from authorization
+ *  @param {number} tokenExp - the time the token expires (UTC)
+ *  @param {string} key - the last key
+ *
+ */
+
+class Servers {
```
blangley28, line 21: I'm also not convinced this needs to be a class. Since there is no `this.server` attribute and each function takes in the `server` as a parameter

src/storage/servers.js
```diff
@@ -0,0 +1,85 @@
...
+
+  getClientId(server) {
+    return this.getServer(server.id).clientId;
+  }
+
+  addAccessToken(server, token) {
+    server.token = token.jwt;
+    server.tokenExp = token.exp;
+    this.addServer(server);
+  }
```
blangley28, line 47: Instead of passing in an object here which doesn't have a defined type I think this should just be
```suggestion
  addAccessToken(server, token, tokenExp) {
    server.token = token;
    server.tokenExp = tokenExp;
    this.addServer(server);
  }
```

src/storage/servers.js
```diff
@@ -0,0 +1,85 @@
...
+
+  async getAccessToken(server) {
+    const result = this.getServer(server.id); // db.select(SERVERS, s => s.id === server.id);
+    if (result.token === undefined || result.tokenExp < Date.now()) {
+      // create a new token if possible
+      try {
+        const jwt = await connectToServer(server.endpoint);
```
blangley28, line 54: This `jwt` is the access token. While Keycloak issues jwts as access tokens it doesn't need to be. I suggest changing this to be `token`

src/utils/client.js
```diff
@@ -1,82 +1,79 @@
...
-      return keystore.get(kid);
-    }
-    return keystore.all({ use: 'sig' });
-  }
+  // Get access token from auth server
+  const data = await axios
+    .post(tokenEndpoint, queryString.stringify(props), headers)
+    .then(response => response.data)
+    .catch(err => console.error(err));
+  return data.access_token;
```
blangley28, line 41: Just a note - here I updated this to return `data.access_token` but you also need the expiration. So I would make this return `data` and then in storage/servers.js get the access_token and expiration. Just as a heads up the expiration is returned in `data.expires_in` which is the seconds until expiration. So when you store it in the db you'll need to add the current date (and convert units since `Date.now()` returns milliseconds)

src/storage/servers.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  addServerKey(server, key) {
+    server.key = key;
+    this.addServer(server);
+  }
+
+  getServerKey(server) {
+    return this.getServer(server.id).key;
+  }
```
blangley28, line 73: What key is this? This app won't know of any server keys. The authorization server (keycloak) maintains all those for us

src/storage/servers.js
```diff
@@ -0,0 +1,108 @@
...
+
+  addServerKeys(server, keys) {
+    keys.serverId = server.id;
+    db.upsert(KEYS, { ...keys }, s => s.id === server.id);
+  }
+
+  getServerKeys(server) {
+    let keys = db.select(KEYS, s => s.serverId === server.id);
+    if (!keys && this.getServerConfiguration(server)) {
+      keys = this.getServerConfiguration(server).jwks;
```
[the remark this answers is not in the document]
dubem7: removed

src/storage/servers.js
dubem7: This returns undefined

src/storage/servers.js
dubem7: This is no longer a class

src/storage/servers.js
dubem7: Updated to match

src/utils/client.js
dubem7: Good call, updated in both places

src/storage/servers.js
dubem7: removed this

src/storage/servers.js
dubem7: changed to token and use the access_token and exp to store in the database

blangley28 (approved):

src/storage/servers.js
```diff
@@ -0,0 +1,96 @@
+const db = require('../storage/DataAccess');
+const { connectToServer } = require('../utils/client');
+
+const SERVERS = 'serverConfig';
```
dehall, line 4: I think the current collection name is just `servers` - let's stick with that

src/storage/servers.js
```diff
@@ -0,0 +1,96 @@
...
+
+function addAccessToken(server, token, tokenExp) {
+  server.token = token;
+  server.tokenExp = tokenExp;
+  addServer(server);
+}
+
+async function getAccessToken(server) {
+  const result = this.getServerById(server.id);
+  if (result.token === undefined || result.tokenExp < Date.now()) {
```
dehall, line 58: Not sure if this actually matters in practice but I see below `clearAccessToken` sets it to null, and here you're checking for undefined. Might be safer to do a `==` instead of `===` here so those two compare equivalently, or just `if (!result.token` to check for falsy values.

src/storage/servers.js
```diff
@@ -0,0 +1,96 @@
...
+    try {
+      const token = await connectToServer(server.endpoint);
+      // expires_in is time until expiration in seconds, Date.now() is in milliseconds
+      const exp = Date.now() + token.expires_in * 1000;
+      db.addAccessToken(server, token.access_token, exp);
+      return token;
+    } catch (e) {
+      clearAccessToken(server);
+      console.error(e);
+      return null;
```
dehall, line 69: So I didn't realize this at first, but after seeing the error in the console from the unit test, I'm wondering if catching the exception and returning null is the right approach here. This is the only place where we call out to the server to get the new token right? So if this returns null that's an exceptional situation, and there's nothing we can do to fix that. I feel like that should stay an exception, still clear the token though, and we should have a handler somewhere higher up that logs it (that handler part would be a separate task).

dehall (approved):

PR #12, opened 2021-02-09
WIP: Reporting workflow

dehall:
Implements the reporting workflow as described at: https://github.com/mcode/medmorph-backend/wiki/Reporting-Workflow

The skeleton is currently in place but there are a number of TODOs still to be addressed.

### DIFF
--- a/src/utils/reporting_workflow.js
+++ b/src/utils/reporting_workflow.js
@@ -1,0 +1,99 @@
+/*
+ * Definition of Reporting Workflow Context:
+ *
+ *
+ */
+
+const IG_REGISTRY = {
+  'http://hl7.org/fhir/us/medmorph/StructureDefinition/us-ph-reporting-bundle' : {} // base action definitions 
+};
+
+function initializeContext(planDefinition) {
+  // TODO
+  return { planDefinition };
+}
+
+function getFunction(ig, code) {
+  // function here in case this needs to be more complex
+  return IG_REGISTRY[ig][code];
+}
+
+function findProfile(planDefinition) {
+  // figure out the profile of the target bundle
+
+  // TODO search arrays rather than hardcode?
+  const createReportAction = planDefinition.action?.find(a => a.code[0].coding[0].code === 'create-report');
+
+  if (!createReportAction) {
+    throw new Error(`PlanDefinition ${planDefinition.id} has no action with code 'create-report'`);
+  }
+
+  const bundleOutput = createReportAction.output?.find(o => o.type === 'Bundle');
+
+  if (!bundleOutput?.profile?.[0]) {
+    throw new Error(`PlanDefinition ${planDefinition.id} does not specify a profile for report bundle`);
+    // option B is just use the default profile
+  }
+
+  return bundleOutput.profile[0];
+}
+
+function determineActionSequence(planDefinition) {
+  // TODO:
+  // note: 6.1.8 
+  // The Backend Service App SHALL implement the ability 
+  // to process action dependencies specified via 
+  // the PlanDefinition.action.relatedAction elements. 
+  // Typical implementations will build a graph or a tree structure 
+  // that can be used to identify dependencies between actions.
+
+
+  // simple version to just get the integers 0 -> n
+  // https://stackoverflow.com/a/10050831
+  return [...Array(planDefinition.action.length).keys()];
+}
+
+
+function executeWorkflow(context) {
+  if (!context.currentActionSequenceStep) {
+    context.currentActionSequenceStep = 0;
+  }
+
+  const planDef = context.planDefinition;
+
+  if (!context.actionSequence) {
+    context.actionSequence = determineActionSequence(planDef);
+  }
+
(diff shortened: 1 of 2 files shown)


src/utils/reporting_workflow.js
```diff
@@ -0,0 +1,154 @@
...
+
+  // simple version to just get the integers 0 -> n
+  // https://stackoverflow.com/a/10050831
+  return [...Array(planDefinition.action.length).keys()];
+}
+
+function executeWorkflow(context) {
+  if (!context.currentActionSequenceStep) {
+    context.currentActionSequenceStep = 0;
+  }
```
blangley28, line 103: Is there a reason this is done here and not in `initializeContext`?

src/utils/reporting_workflow.js
```diff
@@ -0,0 +1,154 @@
...
+
+    const execute = getFunction(ig, actionCode);
+
+    // TODO: maintain context at each step for debugging?
+    // context = deepCopy(context);
+
+    // check the DB to get cancelToken
+    // TODO: is there a better way to do this?
+    const cancelToken = db.select(COLLECTION, c => c.id === context.id).cancelToken;
+    if (cancelToken) return;
```
blangley28, line 131: Does this need to use the db? When would the stored context be different from the context passed to this function?

Also, nothing in the for loop is required for this so it can be moved up to the top of this function.

src/utils/reporting_workflow.js
```diff
@@ -0,0 +1,154 @@
...
+
+    execute(context);
+
+    // update the db after every completed step
```
blangley28, line 135: Is this to support timed events in the future? Everything else should run synchronously right, so the context passed is good enough. And then when a timed event runs it will pull the context from the db and call this method with it?

blangley28 (commented):
Didn't take a deep look at everything because its a WIP but the flow for executeWorkflow looks good.

src/utils/reporting_workflow.js
dehall: Good call - will move that

src/utils/reporting_workflow.js
dehall: The implementation of the cancelToken is admittedly half-baked, at best, and very possibly unnecessary. The idea is that it's something that could be set from elsewhere, in order to terminate this running process. For example, imagine processing one notification for a patient, then a second notification for the same patient comes in right away. Maybe we have the second job check to see if there is a duplicate process running and have it kill the first one, so only one report gets sent with the latest data. So one way to do that is with a cancellation token, which the process checks before every step. So the way for another process to set that is via the database. 

The idea comes from C#/.NET: https://docs.microsoft.com/en-us/dotnet/standard/threading/cancellation-in-managed-threads

src/utils/reporting_workflow.js
dehall: Yeah that's the idea. I didn't think too deep into timed events, but I'm hoping this is a good enough start that we just have to plug a few things into, not necessarily rework the whole thing

blangley28 (approved):
LGTM. One last thing though - since `context` collection isn't a FHIR resource would you mind quickly throwing something up on the [Collections wiki](https://github.com/mcode/medmorph-backend/wiki/Collections)? It looks like its just the context object stored but I think it would be good to just keep a list of all the collections in one place.

src/utils/reporting_workflow.js
```diff
@@ -0,0 +1,195 @@
...
+
+    // TODO: maintain context at each step for debugging?
+    // context = deepCopy(context);
+
+    // check the DB to get cancelToken
+    // TODO: is there a better way to do this?
+    const cancelToken = db.select(COLLECTION, c => c.id === context.id).cancelToken;
+    if (cancelToken) return;
+
+    await execute(context);
```
KeeyanGhoreshi, line 174: is it necessary for this to be await?  If so, do the functions in execute need to be async as well, or will this cover our bases with functions that use promises?

KeeyanGhoreshi (approved):
Looks good to me, though the context is a bit barren.  Is it meant to be a skeleton that can be added onto, or the de-facto structure of the object that all functions should expect?  Some implementations of the actions require more information, like knowledge of the source/destination servers.

blangley28:
> Looks good to me, though the context is a bit barren. Is it meant to be a skeleton that can be added onto, or the de-facto structure of the object that all functions should expect? Some implementations of the actions require more information, like knowledge of the source/destination servers.

@KeeyanGhoreshi that's what my PR is for https://github.com/mcode/medmorph-backend/pull/13

src/utils/reporting_workflow.js
dehall: We need to pick some standard for async operations to make sure things happen in the right order. I figured async here is the easiest, because then the actions that actually need to do async stuff, like call out to a network resource, can just tag themselves as async as well and use await. Not every action needs to be tagged async, this should still work even if they aren't

PR #13, opened 2021-02-12
Initialize Create Report

blangley28:
Links the Subscription Notification to the execute workflow. 
1. Gets Patient and Encounter from the subscription notification
2. Implements the various `context.client` methods using backend auth
3. Refactors `utils/client.js`. A Class is not necessary since there would only ever be one instance. So the jwks can be pulled directly from the privateKey.json file.
4. Modifies the executeWorkflow method to use context.action
5. Updates fetchKA to get Endpoints too

Some open questions:
1. The tests are kind of weak for this. Most of them are just checking the property is set. The client methods aren't tested, should I do this by mocking axios?
2. Right now it fetches all Endpoints instead of just looking at the referenced ones. This was the easiest approach but probably not the most efficient. If this matters, should I fix it here to make a new backlog task?
3. The submit bundle function is using the endpoint defined in the PlanDefinition. For authorization purposes I think this needs to link to Mike's work on persisting authorization details to determine if we have registered with that endpoint. 
4. The resources pulled (right now just Patient and Encounter), are not stored in the database. Should they be added?

Final Comment: Keeyan's PR has not been approved so the final context may change (i.e. see comments about unused properties). This is branched of Dylan's `reporting-workflow` branch should not be merged until Keeyan's `action-implementation` branch is merged.


src/utils/reporting_workflow.js
```diff
@@ -45,13 +50,130 @@ const IG_REGISTRY = {
...
+
   return {
     id: uuidv4(),
+    patient: patient,
+    records: records,
+    encounter: encounter,
+    planDefinition,
+    action: initialAction,
+    messageHeader: null, // Note: not used in actions
+    client: {
```
dehall, line 155: Does the reference SMART on FHIR client support the things we need to do with backend service authorization / custom operations? https://github.com/smart-on-fhir/client-js
If it does then I'd propose we use that rather than invent new things here.

src/utils/reporting_workflow.js
```diff
@@ -45,13 +50,130 @@ const IG_REGISTRY = {
...
+
+  let encounter = null;
+  if (resource.encounter) encounter = await getReferencedResource(resource.encounter.reference);
+  else if (resource.context) encounter = await getReferencedResource(resource.context.reference);
+  else if (resource.resourceType === 'Encounter') encounter = resource;
+
+  // QUESTION: Should encounter and patient be saved to the database?
+
+  const context = initializeContext(planDef, patient, encounter);
+  executeWorkflow(context);
```
dehall, line 128: I wouldn't expect a function called `initialize` to actually start processing. Super minor but can you either rename the function or remove the execute part and just have it return the context?

src/utils/reporting_workflow.js
```diff
@@ -45,13 +50,130 @@ const IG_REGISTRY = {
...
+  if (resource.patient) patient = await getReferencedResource(resource.patient.reference);
+  else if (resource.subject) patient = await getReferencedResource(resource.subject.reference);
+  else if (resource.resourceType === 'Patient') patient = resource;
+
+  let encounter = null;
+  if (resource.encounter) encounter = await getReferencedResource(resource.encounter.reference);
+  else if (resource.context) encounter = await getReferencedResource(resource.context.reference);
+  else if (resource.resourceType === 'Encounter') encounter = resource;
+
+  // QUESTION: Should encounter and patient be saved to the database?
```
dehall, line 125: That's a good question. Since the app in practice would be working with PHI/PII there may be some considerations about storing that.
I also realize that we're storing functions (for the various clients) in the context, which isn't going to work if we want to reload the context. Now I'm thinking maybe we don't want to store the context in the DB at all.

src/utils/reporting_workflow.js
blangley28: How does `startReportingWorkflow` sound?

src/utils/reporting_workflow.js
blangley28: After looking through the docs and a little bit of the code it _might_, but I only see stuff regarding the typical SMART launch not the backend auth. 
Either way I don't really think its necessary since we're not really "inventing" anything. The only extra steps are:
1. Get access token
2. Use access token

The FHIR client definitely supports 2 but adding the authorization header to a a FHIR client call is no easier than attaching it as a header to the axios call.

As for getting the access token, even if the FHIR client did support this for backend auth we already have all of the code implemented to get the access token and Mike is working on caching it. While this isn't necessarily the best defense, we have it and it works so I'm personally hesitant to switch.

src/utils/reporting_workflow.js
blangley28: That's a good point. I can refactor the client functions so they are not included in the context and then update the actions to use those functions. For example, the client.source can just be the endpoint and the submit function defined in action.js

src/utils/reporting_workflow.js
```diff
@@ -45,13 +50,130 @@ const IG_REGISTRY = {
...
+ *
+ * @param {PlanDefinition} planDef - the PlanDefinition resource
+ * @param {*} resource - the resource which triggered the notification
+ */
+async function initializeReportingWorkflow(planDef, resource = null) {
+  // TODO: MEDMORPH-49 to make sure the resource is always included
+  if (!resource) return;
+
+  let patient = null;
+  if (resource.patient) patient = await getReferencedResource(resource.patient.reference);
```
KeeyanGhoreshi, line 116: doesn't this take 2 arguments?  The patient reference will be accepted as the `url` argument

src/utils/fhir.js
```diff
@@ -37,21 +38,30 @@ function generateOperationOutcome(code, msg) {
...
+ * @param {string} resourceType - the type of the resource
  */
-function getResources(server, type, ids = null) {
-  const id_param = ids ? `&_id=${ids.join(',')}` : '';
-  const url = `${server}/${type}?_include=*${id_param}`;
-  const headers = { Authorization: 'Bearer admin' }; // TODO: get an access token
-  return axios
+async function getResources(server, resourceType, db) {
+  const url = `${server}/${resourceType}?_include=*`;
+  const token = await connectToServer(server);
```
KeeyanGhoreshi, line 49: There seems to be several calls to `connectToServer`, and it asks for another access_token each time.  Is there a way we can just use the same token or refresh the token if the URL is the same?

KeeyanGhoreshi (changes requested):

src/utils/reporting_workflow.js
```diff
@@ -45,13 +50,130 @@ const IG_REGISTRY = {
...
-function initializeContext(planDefinition) {
-  // TODO -- see MEDMORPH-36
+/**
+ * Retrieve data from the EHR
+ *
+ * @param {string} uri - the search query for the data desired
+ * @returns axios promise with data
+ */
+async function readFromEHR(uri) {
+  const url = process.env.EHR;
```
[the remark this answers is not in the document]
blangley28: So originally I made it an env var because there would only be one EHR, but then Dylan pointed out it still needs to be in the servers collection anyway to store the authorization details. I decided against removing the env var but I'm thinking now there's no point to having it

src/utils/fhir.js
blangley28: That's Mike's PR #11

src/utils/reporting_workflow.js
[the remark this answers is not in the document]
blangley28: Once Mike's PR is merged I will add on to this storage/servers interface to implement a `getEHRServer()` method

blangley28:
I'm not sure how to mock the readFromEHR or submitBundle methods so I can't update the tests for extract-research-data nor submit-report actions.

src/utils/reporting_workflow.js
blangley28: Refactored the client methods but haven't done anything regarding storing the encounter or patient.

blangley28:
In the latest commit I was able to get the entire flow working. On the notification of a Subscription it will correctly go through all of the actions. I created a PlanDefinition (see Knowledge Artifact > Requests > Upload PlanDefinition https://www.getpostman.com/collections/de9729029bae515d3368)

src/utils/actions.js
```diff
@@ -166,58 +165,64 @@ const baseIgActions = {
...
+              filteredResources = filteredResources.filter(resource => {
+                const code = getBySearchParam(resource, searchParam);
+                return compareCodes(code, codeFilter.code);
+              });
+            }
           }
-        }
-        return context.records.push(...filteredResources);
+
+          // QUESTION - should context have a mutex?
```
dehall, line 219: I don't think it's necessary at this point. We aren't expecting to have multiple processes using the same object. If there are going to be 2 things looking at the same context it will be via the DB so not the same actual object

src/utils/actions.js
```diff
@@ -76,12 +80,13 @@ const baseIgActions = {
...
   'submit-report': async context => {
-    await context.client.dest
-      .submit(context.reportingBundle)
+    await submitBundle(context.reportingBundle, context.client.dest)
       .then(
         result => {
+          console.log(result);
           if (result.status === StatusCodes.ACCEPTED || result.status === StatusCodes.OK) {
             context.flags['submitted'] = true;
+            debug(`Reporting Bundle submitted to ${context.client.dest}`);
```
dehall, line 89: I like the `debug` lines better than the console.log/error but thinking forward a little bit to real use, can you add a bundle identifier of some kind to this line, so we can tell which bundle was submitted? Don't worry about adding new debug lines anywhere you haven't already, but where you do have them please add an ID for traceability

README.md
```diff
@@ -22,14 +22,18 @@ npm start
...
+Certain configuration properties MUST be set via enviornment variables. The easiest way to do this is to set them in an environment file, `.env`. A sample file is provided in this repo as `.env.dev` (you will need to rename/copy this into the `.env` file for it to work).
 
 | ENV | Required | Description |
 | --- | -------- | ----------- |
 | REALM | Yes | The Keycloak realm name. This is the url name which may be different than the text name. |
 | AUTH | Yes | The Keycloak auth url. Will look something like `http://keycloak.example.com/auth`/ |
 | PORT | No | The port number for the server. Defaults to 3000 if not provided. |
 | DEBUG | No | Set to `medmorph-backend:*` to enable debug loggers for the app. |
+| DATA_TRUST_SERVICE | Yes | The base url for the data/trust service server. |
+| BASE_URL | Yes | The base url for this server. |
```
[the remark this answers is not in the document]
blangley28: Its from Keeyan's action PR. Its used to populate the MessageHeader.source.endpoint

src/utils/actions.js
blangley28: Good point. Removing this comment

dehall:
I noticed this in the console when running tests locally:
> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.

The tests all pass but that message didn't show up on `main`. (And I tried the --detectOpenHandles thing but it did nothing)

blangley28:
@dehall so I have figured out that is because of the `setTimeout` I added to `app.js`. I've tried using `jest.useFakeTimers()` and `global.setTimeout = jest.fn()` but neither of them work. I'll have to do some more digging next week as to how to fix it

KeeyanGhoreshi (approved):

dehall (approved):

PR #14, opened 2021-02-17
Make jwks url public

blangley28:
Quick fix to make the JWKs url not require backend auth. Created a new /public route for endpoints which don't require authorization


danlee1025 (approved):
LGTM

PR #15, opened 2021-02-18
Handle offsets in the reporting workflow

dehall:
PlanDefinitions in the MedMorph IG allow for timed offsets between actions, using the `PlanDefinition.action.relatedAction.offsetDuration` field.

This PR adds support for these timed offsets, by treating an offset as a new entry in the action sequence that causes a break out of the main loop, and calls setTimeout with the context pointing to the next action. It's admittedly very hackish but I feel like any more elegant solution will be a lot longer. I'm open to suggestions.

Note that this does not add a job scheduler because I'm not entirely sure we need one for this. The place it would be useful is if there are long delays where a system restart may happen in the middle, but beyond that I don't think a job scheduler adds much benefit. Again I'm open to hearing other opinions.

### DIFF
--- a/src/utils/__tests__/reporting_workflow.test.js
+++ b/src/utils/__tests__/reporting_workflow.test.js
@@ -227,4 +227,35 @@ describe('Test reporting workflow', () => {
     // G must be before I
     expect(g).toBeLessThan(i);
   });
+
+  test('execution works with offsets between actions', done => {
+    // setup a new action specifically for this test
+    EXAMPLE_ACTION_IMPL['do-something-offset'] = () => done();
+
+    const planDef = {
+      resourceType: 'PlanDefinition',
+      meta: {
+        profile: [MEDMORPH_PLANDEF_PROFILE]
+      },
+      action: [
+        buildAction({ code: 'init-counter', next: 'increment-counter' }),
+        buildAction({ code: 'increment-counter', next: 'do-something-offset' }),
+        buildAction({ code: 'do-something-offset', next: 'create-report' }),
+        CREATE_REPORT_ACTION
+      ]
+    };
+
+    const duration = {
+      unit: 's',
+      value: 1
+    };
+
+    planDef.action[0].relatedAction[0].offsetDuration = duration;
+    planDef.action[1].relatedAction[0].offsetDuration = duration;
+
+    const context = initializeContext(planDef);
+    executeWorkflow(context);
+    // the workflow will run and call done() async.
+    // if you don't believe me, change the action impl above to not call done()
+  });
 });
--- a/src/utils/__tests__/reporting_workflow.test.js
+++ b/src/utils/__tests__/reporting_workflow.test.js
@@ -246,8 +246,8 @@ describe('Test reporting workflow', () => {
     };
 
     const duration = {
-      code: 's',
-      value: 1
+      code: 'ms',
+      value: 50
     };
 
     planDef.action[0].relatedAction[0].offsetDuration = duration;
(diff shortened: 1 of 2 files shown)


src/utils/reporting_workflow.js
```diff
@@ -131,11 +131,28 @@ function determineActionSequence(planDefinition) {
...
-        // directed edge from arg1 --> arg2
-        // TODO: track the offset on the edge label maybe?
-        g.setEdge(action.id, relatedAction.actionId);
-        // note that relatedAction.id doesn't already have to be in the graph
-        // it will be added if not
+        if (relatedAction.offsetDuration) {
+          // assume that the offsetDuration indicates a minimum, not a maximum.
+          // Duration object has a 'comparator' field which can be </<=/>/>=
+          // TODO: if there's a case of offset < x, just ignore it and run immediately
+          const { value, unit } = relatedAction.offsetDuration;
```
blangley28, line 205: In Duration it looks the code is used. I haven't worked with Duration before the example PlanDefinition has the code as "min" instead of unit as "min". The code system must be UCUM so we don't need to worry about different systems. http://hl7.org/fhir/R4/datatypes.html#Duration
```suggestion
          const { value, code } = relatedAction.offsetDuration;
```

src/utils/reporting_workflow.js
```diff
@@ -218,18 +233,54 @@ function determineActionSequence(planDefinition) {
...
 
   while (context.currentActionSequenceStep < context.actionSequence.length) {
+    const currentActionId = context.actionSequence[context.currentActionSequenceStep];
+
+    if (currentActionId.startsWith('&&')) {
+      // delay the specified amount of time
+      const waitTime = Number(currentActionId.substring(2).split(' ')[0]); // split off the first && and pick out the time
+      context.currentActionSequenceStep++;
+
+      // TODO: do we even need a job scheduler?
```
blangley28, line 272: From some online literature it appears long setTimeouts aren't an issue. However, the only thing to consider is if we delay by a week and the server needs to be restarted the timeout thread would be lost. I'm also not sure how well it scales. For current purposes I think this is fine though

src/utils/__tests__/reporting_workflow.test.js
```diff
@@ -227,4 +227,35 @@ describe('Test reporting workflow', () => {
...
+      value: 1
+    };
+
+    planDef.action[0].relatedAction[0].offsetDuration = duration;
+    planDef.action[1].relatedAction[0].offsetDuration = duration;
+
+    const context = initializeContext(planDef);
+    executeWorkflow(context);
+    // the workflow will run and call done() async.
+    // if you don't believe me, change the action impl above to not call done()
```
blangley28, line 259: Hahaha, I believe you. One thing to note is the this test then delays and runs slowly. I believe the solution should be `jest.useFakeTimers()`, but I failed to get that working for my PR and adding it to this test didn't work either. Something to consider though

blangley28 (commented):
Looks good, I like the approach. Just a minor change and some comments

src/utils/__tests__/reporting_workflow.test.js
dehall: Yeah that's why I intentionally made it only 2 seconds total so it's not a huge delay, but I'll look into useFakeTimers

blangley28: Can we make it like 50ms or at least something shorter?

dehall: Updated to 50ms

blangley28 (approved):

danlee1025 (approved):
This looks good to me.

PR #16, opened 2021-02-22
Shell script for adding server to collection

dehall:
Adds a quick shell script for adding a server to the servers collection. 

Unfortunately there's no good way to make this work both when the app is running or not running. The issue is that LokiJS doesn't support two processes writing to the same db file, so if you try to do that you will almost certainly cause data loss. Another possibility would be to first check if the server is running, and if not add it to the collection directly, otherwise POST to the /servers endpoint. That feels like overkill but I'm willing to explore it.

Run this script either directly with `./bin/add-server.sh -e endpoint [-n name] [-t type] [-c clientId]` or via npm with `npm run add-server -- -e endpoint [-n name] [-t type] [-c clientId]` (note the extra `--` is needed to tell npm the remaining args are for the command, not npm itself)

Eventually this will likely be OBE once we add an admin UI, but even still this will be a very quick way to add a server to the collection.


dehall:
Attached is an alternative implementation, mostly complete, that checks to see if the app is running - if it is, use the API, if not, go to the DB directly. It's a little cleaner, because it's JS vs shell, but it still introduces the possibility of data loss if it fails to recognize that the app is running somehow. Any thoughts or preferences from reviewers?

[add-server.txt](https://github.com/mcode/medmorph-backend/files/6044764/add-server.txt)

KeeyanGhoreshi:
Personally I prefer the JS, since I'm more familiar with it than the shell script, but either way if this is just an interim for until we get the admin UI up and running I'm not too concerned with the specifics.

KeeyanGhoreshi (approved):
I think it looks fine as an interim script.

blangley28:
I can't get this to run. I noticed the comment regarding addToDb so I'm running the server and trying to set a new server via callApi. Does this look correct 
```
npm run add-server-js -e http://example.org -t EHR -n Test
```

dehall:
You'll need to add `--` to tell npm that any subsequent args should be sent to the sub-command rather than parsed by npm itself.
i.e. `npm run add-server-js -- -e http://example.org -t EHR -n Test`

bin/add-server
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const baseURL = process.env.BASE_URL;
+
+// see if the server is already running
+// by hitting the base url.
+// no connection = it's not running, add to DB directly
+// connection succeeds = it's running, use the API
+
+axios.get(baseURL).then(callApi).catch(addToDb);
```
blangley28, line 61: This always calls `addToDb` for me because the baseUrl requires authentication. The two solutions are to 
1. Change this to get a public endpoint (i.e. `/public/jwks`) or,
2. Add the admin token in an authorization header

bin/add-server
```diff
@@ -0,0 +1,87 @@
...
+  // TODO: this doesn't actually work. need to async wait for it to init
+  // and even then it seems to hang for no reason
+  servers.addServer(newServer);
+  console.log('Added successfully.');
+}
+
+function callApi() {
+  try {
+    console.log('App running. Adding server via api.');
+    const addServerUrl = baseURL + (baseURL.endsWith('/') ? '' : '/') + 'servers';
```
blangley28, line 76: We don't do this anywhere else in the app (it is expecting the baseUrl to not have a `/` at the end). I feel like we're getting to the point where a config manager might be required so we can call something like `config.getBaseUrl()` and it will handle this logic. I know we've discussed config stuff for a bit now

bin/add-server
```diff
@@ -0,0 +1,87 @@
...
+axios.get(baseURL).then(callApi).catch(addToDb);
+
+function addToDb() {
+  console.log('App not running. Adding server to db directly.');
+
+  const servers = require('../src/storage/servers');
+  // TODO: this doesn't actually work. need to async wait for it to init
+  // and even then it seems to hang for no reason
+  servers.addServer(newServer);
+  console.log('Added successfully.');
```
blangley28, line 70: So you need to wait for to initialize but then it hangs because the db is still running. I couldn't find a way to have loki shut down so using `process.exit(0)` did the trick... but you also have to wait for `servers.addServer()` to complete before calling `exit`. This is not a pretty solution but it worked for me 
```
const servers = require('../src/storage/servers');
  setTimeout(() => {
    servers.addServer(newServer);
    setTimeout(() => {
      console.log('Added successfully.');
      process.exit(0);
    }, 10000);
  }, 10000);
```

blangley28 (changes requested):

bin/add-server
dehall: If there's an expectation that the baseUrl won't have a `/` then I'll continue with that

bin/add-server
dehall: Yeah this is where that idea I had of refactoring the DB post-init stuff would be useful. Not sure if I want to do something like this first, then refactor later, or just go for it and refactor things here

blangley28: I'm good to refactor it later. Created [MEDMORPH-77](https://jira.mitre.org/browse/MEDMORPH-77).

dehall: Too late I did it here. It's only like 10 lines.

dehall:
In the latest push I just added the "postinit" file to help with running things only after the DB is ready. It clarifies the intent of the setTimeouts in app.js, and should remove the warning message we were seeing on tests too

bin/add-server
```diff
@@ -0,0 +1,98 @@
...
+  const servers = require('../src/storage/servers');
+
+  runWhenDBReady(() => {
+    servers.addServer(newServer);
+
+    // TODO: ideally we wouldn't have to end the process manually
+    setTimeout(() => {
+      console.log('Added successfully.');
+      process.exit(0);
+    }, 6000);
```
blangley28, line 81: This worked the first time for me but not the second. I'm wondering if instead of always saying "Added successfully" it should query the database for that server and see if its returned, then change the message.

bin/add-server
```diff
@@ -0,0 +1,98 @@
...
+}).then(callApi).catch(addToDb);
+
+function addToDb() {
+  console.log('App not running. Adding server to db directly.');
+
+  const { runWhenDBReady } = require('../src/storage/postinit');
+  const servers = require('../src/storage/servers');
+
+  runWhenDBReady(() => {
+    servers.addServer(newServer);
```
blangley28, line 75: When adding the server this way we loose out on the `id` field.... perhaps its best to have the `addServer` method check for an `id` and add one if it doesn't already exist on the server.

bin/add-server
```diff
@@ -0,0 +1,98 @@
...
+      console.error(`Unknown command line option: ${argKey}`);
+      console.error(`Full args: ${args}`);
+      process.exit(1);
+  }
+}
+
+if (!newServer['endpoint']) {
+  console.error("Required argument -e|--endpoint missing!");
+  process.exit(1);
+}
```
blangley28, line 46: The required fields also include `type` and `clientId`. The app won't work correctly without them

blangley28 (changes requested):

bin/add-server
dehall: I'll move the "add ID if not present" logic from the /servers route to the addServer method. This will also fix the issue with it seemingly not working when you run it twice

bin/add-server
dehall: So in this case that actually wouldn't have helped. What's going on here is the `addServer` function does an `upsert` on the server with the same ID, so if we already had an entry with no id, it will just update that and not insert a new one. I'm assuming that if it does actually fail to do anything, it will throw an exception

blangley28 (approved):

PR #17, opened 2021-02-23
add frontend

KeeyanGhoreshi:
The endpoints were changed for the backend to include a /api, which is only really relevant for the index router but for consistency was extended to the other endpoints, so that the frontend isn't overlapping with any of the backend endpoints.

The npm-run-all package lets us run both the servers with one command for development.  For deployment, the app has a line that hosts the static files when they're built out, and does nothing when the build folder doesn't exist, which lets the express app host the files itself so we don't have two servers running unless its for development. 

When both servers are running, the frontend proxies requests it can't service to the backend automatically, which gives the illusion of only one server running.  This gives us all the benefits of development tools like hot reloading when working on the frontend/backend individually, while still running both.


client/src/components/App.css
```diff
@@ -0,0 +1,7 @@
+.App {
```
blangley28, line 1: Do we want to use the same approach as Pathways using themes and `useStyles()`

src/app.js
```diff
@@ -39,9 +39,9 @@ app.use('/.well-known', wellKnownRouter);
 app.use('/public', publicRouter);
 
 // Protected Routes
-app.use('/', backendAuthorization, indexRouter);
-app.use('/fhir', backendAuthorization, fhirRouter);
-app.use('/servers', backendAuthorization, serversRouter);
-app.use('/notif', subscriptionAuthorization, subscriptionsRouter);
+app.use('/api',backendAuthorization, indexRouter);
```
blangley28, line 42: Can you be sure to update the wiki with these new endpoints please

blangley28 (commented):
I can't actually get this to run. Should the steps just be
1. `cd client`
2. `npm install`
3.  `cd ..`
4. `npm start`

KeeyanGhoreshi:
The steps you posted are correct, including an `npm install` in the main folder to ensure you have the `npm-run-all` package.

src/app.js
KeeyanGhoreshi: i decided to revert some of the `/api` endpoints, since they didn't make sense, but I'll update the index one.

blangley28 (approved):

client/README.md
```diff
@@ -0,0 +1,3 @@
+# Running the frontend
+
+Run `npm start` in the client directory to run the frontend by itself.  Running `npm start` in the parent directory should also start the frontend server as well.  Requests that the client server cannot service will be automatically proxied to the backend.
```
dehall, line 3: Can you clarify what "requests that the client server cannot service will be automatically proxied to the backend" means? I don't see anything that seems like it would do that.

dehall (commented):
A few thoughts:
 - Is it possible to even include the UI within the same base project as the backend? Or is that too crazy/not possible/not recommended for reasons x/y/z? Seems like since it's all javascript we could have it all in one place and not have to deal with multiple projects.
 - Or if we keep it the same structure, can you rename the folder from `client` to `ui`? `client` is a little ambiguous because this whole thing is also the "backend services client"
 - I guess to run this in production we would npm build && npm run startBackend? Can you add a note to the top-level README?

client/README.md
KeeyanGhoreshi: client package.json includes a "proxy" property that does this.

KeeyanGhoreshi:
It is possible to move the client/UI folder elsewhere, but the basic workflow will be the same no matter what.  In development, we proxy requests from the frontend to the backend, so that the servers are separate and can be developed individually.  In production the frontend is built out into static files that the backend serves.  

Personally I don't have a preference for how the client is contained in the project, most of what I read pointed towards making the react app its own project within the express app to make it easier to handle.  I'm not 100% sure how we'd get it to work making it part of the same project.

dehall:
Ok thanks. Let's keep it like this then. Building static files every time the server starts is pretty slow and I'm not sure it would allow for hot-deploying when things change.
In that case can you just rename the folder to `ui`?

KeeyanGhoreshi:
hold on im overhauling the entire thing

KeeyanGhoreshi:
so I changed it to only be one project, sharing a package.json, using webpack to compile and watch the react files.  This setup, I think, is a lot less confusing than the previous one.  Webpack builds out the react files each time.

Pros:
- less confusing
- all on project
- no management of two package.json and two .gitignores etc
- behaves the same way in production

Cons: 
- it's slightly slower than two separate servers

package.json
```diff
@@ -5,20 +5,31 @@
...
   "description": "MedMorph Backend Service Client",
   "scripts": {
-    "start": "nodemon ./bin/www",
+    "start": "npm-run-all --parallel startFrontend startBackend",
+    "startBackend": "nodemon ./bin/www",
+    "startFrontend": "webpack",
     "debug": "node --inspect-brk ./bin/www",
     "test": "jest",
     "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand",
     "lint": "eslint './src/**/*.{js,jsx}'",
```
dehall, line 14: Can you make sure that `lint` and `test` work for both the ui and backend? Seems like test should work without any addtl changes, lint looks like it will need the `./ui/` folder added to args. Also we could theoretically add the ui folder under src right?

KeeyanGhoreshi: moving ui under src also makes it part of the lint, so that works out

blangley28 (commented):
Changes look good to me. Only weird thing is that when I run the express server starts, then the frontend, and then the express server restarts. It looks like perhaps the frontend starting is changing a file causing the the backend server to restart due to changes. Seems like just a development thing and its not the end of the world but is kind of weird.

KeeyanGhoreshi:
it's because the express server is watching for changes to the static js files, and webpack is looking for changes to the .jsx files to update the static js files.  When running the `start` command it will start both in parallel, so what happens is the `public/js` folder is empty, and the express backend spins up before the `bundle.js` file is made/updated.  By the time webpack has compiled the files and created `bundle.js`, the express server is already running and checking for changes, so it reloads.  

The "solution" is to just run the frontend first to compile the files first.  It's just a quirk of the `start` command starting both servers in parallel to circumvent the hassle of having to manually start both.

dehall (approved):

PR #18, opened 2021-02-24
Forward message response to EHR

blangley28:
Forward message response (either from an async post to $process-message or the sync response from submitting the bundle) to the EHR server.


KeeyanGhoreshi (approved):
Looks good to me.

PR #19, opened 2021-02-24
Knowledge Artifact Subscriptions

blangley28:
On startup subscribe to PlanDefinition on all KA repos. Implement the subscription notification endpoints to store the updated artifacts on change.

HAPI won't let me subscribe to criteria `PlanDefinition` so I opted for `PlanDefinition?_lastUpdated=gt2021-01-01`. If anyone knows a better way to subscribe to all PlanDefinitions without requiring the R5 backport topics let me know.

PR #20, opened 2021-02-25
Subscription setup

danlee1025:
Currently this is branched off of Ben's PR #19 to incorporate the subscription changes there.
When the PlanDefinitions from KA are refreshed, subscriptions are saved to the database and created on the EHR server.

- Call `subscriptionsFromBundle` whenever PlanDefinitions are refreshed.
- Added an id to subscription resources that is `${planDef.id}${namedEventCoding.code}` so that they can be updated in the database.
    - Using `axios.put` so that the EHR server uses this `id` so it will know when to create/update.
- Added a regex in `generateSubscription` to check if `criteria` has at least one search param since HAPI is expecting this(mentioned in #19).
    - If `criteria` doesn't have a search param, I add `?_lastUpdated=gt2021-01-01` to it so we can post to EHR server. If we come to a different solution in #19 I'll change it here also.

### DIFF
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -153,6 +153,7 @@ function refreshAllKnowledgeArtifacts() {
 
 /**
  * Get knowledge artifacts from a specific server.
+ * Generate Subscription resources from the KA Bundle with PlanDefinition
  *
  * @param {*} server - the server to refresh artifacts from
  */
@@ -188,6 +189,7 @@ function subscribeToKnowledgeArtifacts() {
 
 /**
  * Take a Knowledge Artifact Bundle and generate Subscription resources for the named events
+ * Saves Subscription resources to DB and posts them to EHR server
  *
  * @param {Bundle} specBundle - KA Bundle with PlanDefinition
  * @param {string} url - the notification endpoint url
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -48,7 +48,7 @@ function generateOperationOutcome(code, msg) {
  *
  * @param {string} server - the sourse server base url
  * @param {string} resourceType - the type of the resource
- * @returns FHIR bundle of resources
+ * @returns axios promise of FHIR bundle of resources
  */
 async function getResources(server, resourceType) {
   const url = `${server}/${resourceType}?_include=*`;
@@ -205,17 +205,19 @@ function subscriptionsFromBundle(specBundle, url, token) {
     .flat();
 
   subscriptions.forEach(async subscription => {
+    const subscriptionId = subscription.id;
+
     // Store subscriptions in database
-    debug(`  ...${subscription.resourceType}/${subscription.id}`);
-    db.upsert('subscriptions', subscription, s => s.id === subscription.id);
+    debug(`  ...Subscription/${subscriptionId}`);
+    db.upsert('subscriptions', subscription, s => s.id === subscriptionId);
 
     // Create/Update Subscriptions on EHR server
     const ehrServer = getEHRServer();
     const ehrToken = await getAccessToken(ehrServer.endpoint);
     const headers = { Authorization: `Bearer ${ehrToken}` };
     axios
-      .post(`${ehrServer.endpoint}/Subscription`, subscription, { headers: headers })
-      .then(() => debug(`Subscription created on EHR server`));
+      .put(`${ehrServer.endpoint}/Subscription/${subscriptionId}`, subscription, { headers })
+      .then(() => debug(`Subscription with id ${subscriptionId} created/updated on EHR server`));
   });
 
   return subscriptions;
@@ -241,7 +243,7 @@ function subscriptionsFromPlanDef(planDef, url, token) {
     );
 
     const criteria = namedEventToCriteria(namedEventCoding.code);
-    const id = `${planDef.id}${namedEventCoding.code}`;
+    const id = `sub${planDef.id}${namedEventCoding.code}`;
     if (criteria)
       subscriptions.push(generateSubscription(criteria, url, token, id, namedEventCoding.code));
     return subscriptions;
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -213,11 +213,13 @@ function subscriptionsFromBundle(specBundle, url, token) {
 
     // Create/Update Subscriptions on EHR server
     const ehrServer = getEHRServer();
-    const ehrToken = await getAccessToken(ehrServer.endpoint);
-    const headers = { Authorization: `Bearer ${ehrToken}` };
-    axios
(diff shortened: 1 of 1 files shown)


src/utils/fhir.js
```diff
@@ -186,7 +196,25 @@ function subscriptionsFromBundle(specBundle, url, token) {
...
+  const subscriptions = planDefs
+    .map(planDef => subscriptionsFromPlanDef(planDef.resource, url, token))
+    .flat();
+
+  subscriptions.forEach(async subscription => {
+    // Store subscriptions in database
+    debug(`  ...${subscription.resourceType}/${subscription.id}`);
+    db.upsert('subscriptions', subscription, s => s.id === subscription.id);
+
+    // Create/Update Subscriptions on EHR server
```
blangley28, line 208: Just to clarify - due to the open question you have on this PR right now this only creates, it doesn't update. I just tested it out and using PUT will handle the create/update for us and it will still handle has a normal subscription (i.e. send notifications correctly. I think the easiest solution then is to just make this `axios.put` to `{ehr}/Subscription/{subscription id}`. For HAPI, we need to make sure the id starts with a letter so if we go this route I would change line 240 to be 
```
const id = `sub${planDef.id}${namedEventCoding.code}`;
```

blangley28 (changes requested):

src/utils/fhir.js
danlee1025: I like this approach too instead of trying to keep track multiple ids. Updated the code to use this!

src/utils/fhir.js
```diff
@@ -76,10 +80,13 @@ async function getResources(server, resourceType) {
  * @param {string} subscriptionTopic - R5 backport subscription topic
  * @returns a R5 Backport Subscription
  */
-function generateSubscription(criteria, url, token, subscriptionTopic = undefined) {
+function generateSubscription(criteria, url, token, subscriptionTopic = undefined, id = undefined) {
+  // Add regex to check if criteria has at least one search param since HAPI is expecting this
+  const criteriaReg = /(.+)(\?.)+/;
   const subscription = {
+    ...(id && { id }),
```
blangley28, line 87: Thoughts on making this:
```suggestion
    id: id ?? shortId.generate(),
```

blangley28: Or since we already have uuid as a package we could 
```
const { v4: uuidv4 } = require('uuid');
...
id: id ?? uuidv4(),
```

danlee1025: Made this change ``id: id ?? `sub${uuidv4()}`,`` in case we wanted to pass the id to HAPI.

blangley28:
Code looks good. When I run `npm test` I see a whole bunch of output saying 
> UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'endpoint' of undefined

which doesn't show up on the `ka-subscriptions` branch. Could you take a look at why that's happening?

danlee1025:
Ahh one of the tests call `subscriptionsFromBundle` and the `ehrServer` is not available in the test so I added a null check before posting the resource.

src/utils/fhir.js
```diff
@@ -186,7 +197,29 @@ function subscriptionsFromBundle(specBundle, url, token) {
...
-  return planDefs.map(planDef => subscriptionsFromPlanDef(planDef.resource, url, token)).flat();
+  const subscriptions = planDefs
+    .map(planDef => subscriptionsFromPlanDef(planDef.resource, url, token))
+    .flat();
+
+  subscriptions.forEach(async subscription => {
+    const subscriptionId = subscription.id;
+
+    // Store subscriptions in database
+    debug(`  ...Subscription/${subscriptionId}`);
```
blangley28, line 208: One more thing, can this be restructured. It makes it look like this Subscription was fetched from the server but it wasn't

```
  medmorph-backend:server Fetched http://localhost:8090/fhir/Bundle/7f06514d-7d21-4635-b488-82fc2aa7af76 +184ms
  medmorph-backend:server   ...PlanDefinition/PD1 +0ms
  medmorph-backend:server   ...PlanDefinition/plandefinition-cancer-example +0ms
  medmorph-backend:server   ...Subscription/subPD1encounter-close +1ms
  medmorph-backend:server   ...Subscription/subplandefinition-cancer-exampleencounter-close
```

danlee1025: I can change it to something like `Saved Subscription with id ${subscriptionId} to db.` or should I just remove the debug statement since there's one right below for posting to ehr server?

blangley28: I would keep the debug because its nice to have the log. Your suggestion sounds good, or even just replace `  ...` with `Saved `

src/utils/fhir.js
```diff
@@ -152,7 +157,10 @@ function refreshAllKnowledgeArtifacts() {
  * @param {*} server - the server to refresh artifacts from
  */
 function refreshKnowledgeArtifact(server) {
-  getResources(server.endpoint, 'PlanDefinition');
+  getResources(server.endpoint, 'PlanDefinition').then(bundle =>
+    // Create/Update subscriptions from PlanDefinitions
+    subscriptionsFromBundle(bundle, server.endpoint, 'admin')
```
blangley28, line 162: The original intent of this function was just to create the subscription resources. Since it now does more can you update the description of the function please

blangley28 (approved):
LGTM. Just that last minor comment about updating a commnet

src/utils/fhir.js
```diff
@@ -80,10 +83,12 @@ async function getResources(server, resourceType) {
  * @returns a R5 Backport Subscription
  */
 function generateSubscription(criteria, url, token, id = undefined, subscriptionTopic = undefined) {
+  // Add regex to check if criteria has at least one search param since HAPI is expecting this
+  const criteriaReg = /(.+)(\?.)+/;
   const subscription = {
     id: id ?? `sub${uuidv4()}`,
     resourceType: 'Subscription',
-    criteria: `${criteria}`,
+    criteria: criteriaReg.test(criteria) ? `${criteria}` : `${criteria}?_lastUpdated=gt2021-01-01`,
```
KeeyanGhoreshi, line 91: is the `_lastUpdated` param necessary?  I think just a question mark will suffice.

KeeyanGhoreshi (commented):
Seems good,

src/utils/fhir.js
blangley28: You're right it isn't. I just tested and the Subscription is accepted as long as it has the `?` and the notification still goes out correctly.

danlee1025: Updated regex to check for just the question and removed `_lastUpdated`

KeeyanGhoreshi (approved):

PR #21, opened 2021-02-25
Add authorization to subscriptions

blangley28:
Add authorization to Subscription notifications. Generates an access token to be included in the auth header. Also implements the authorization for validating that token. I took one approach to reading it from the Subscription resource but I'm open to other methods.

I also added `storage/collections.js` to define constants for the collections in the db. It was getting frustrating to define these in each file or as strings (I'm prone to typos).

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
   "dependencies": {
     "axios": "^0.21.1",
     "babel-eslint": "^10.1.0",
-    "crypto": "^1.0.1",
     "debug": "~2.6.9",
     "dotenv": "^8.2.0",
     "eslint": "^7.19.0",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "dependencies": {
     "axios": "^0.21.1",
     "babel-eslint": "^10.1.0",
+    "crypto": "^1.0.1",
     "debug": "~2.6.9",
     "dotenv": "^8.2.0",
     "eslint": "^7.19.0",
--- a/src/utils/auth.js
+++ b/src/utils/auth.js
@@ -1,4 +1,8 @@
+const crypto = require('crypto');
+const { StatusCodes } = require('http-status-codes');
 const passport = require('passport');
+const { SUBSCRIPTIONS } = require('../storage/collections');
+const db = require('../storage/DataAccess');
 
 /**
  * Helper function to obtain the token from the authorization header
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
   "dependencies": {
     "axios": "^0.21.1",
     "babel-eslint": "^10.1.0",
+    "crypto": "^1.0.1",
```
dehall, line 38: Is this https://www.npmjs.com/package/crypto ? That page says this library is deprecated and built into node now.

dehall (approved):

src/utils/auth.js
```diff
@@ -25,9 +38,27 @@ function backendAuthorization(req, res, next) {
...
+        // Get the authorization token from the subscription
+        const authorizationHeader = subscription[0].channel.header[0];
+        const authToken = authorizationHeader.split('Authorization: Bearer ')[1];
+        if (token === authToken) return next();
+      }
+    }
+  }
+
+  res.sendStatus(StatusCodes.UNAUTHORIZED);
+  return;
```
danlee1025, line 61: Can this `return` statement be removed? Function is going to end anyway right?

src/utils/fhir.js
```diff
@@ -77,24 +79,24 @@ async function getResources(server, resourceType) {
...
- * @param {string} token - access token for header
  * @param {string} id - the id to assign the Subscription
  * @param {string} subscriptionTopic - R5 backport subscription topic
  * @returns a R5 Backport Subscription
  */
-function generateSubscription(criteria, url, token, id = undefined, subscriptionTopic = undefined) {
+function generateSubscription(criteria, url, id = undefined, subscriptionTopic = undefined) {
   // Add regex to check if criteria has at least one search param since HAPI is expecting this
-  const criteriaReg = /(.+)(\?)+/;
+  const criteriaReg = /(.+)(\?.)+/;
```
danlee1025, line 88: Might just be a rebase error but I removed the `.` from the regex so it only checks for the question mark.
```suggestion
  const criteriaReg = /(.+)(\?)+/;
```

src/utils/auth.js
blangley28: Yea you're right. Originally it wasn't at the end so it was needed but then I refactored the branching. I will remove it

danlee1025 (approved):

PR #22, opened 2021-02-26
dockerize backend

KeeyanGhoreshi:
Adds a dockerfile and docker-compose, which lets us launch the app with docker.  Very basic implementation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,4 @@ COPY package*.json ./
 RUN npm install
 COPY . .
 EXPOSE 3000
-CMD npm start
+CMD npm run-script startProd
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:14
+
+# Create app directory
+WORKDIR /usr/src/app
+COPY package*.json ./
+RUN npm install
+COPY . .
+EXPOSE 3000
+CMD npm start
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,13 @@
+version: "3"
+
+services:
+  medmorph_backend:
+    image: medmorph_backend
+    container_name: medmorph_backend
+    restart: on-failure
+    ports:
+      - "3000:3000"
+    environment:
+      - AUTH=http://moonshot-dev.mitre.org:8090/auth
+      - REALM=backend_app
+      - BASE_URL=http://localhost:3000
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,3 +11,4 @@ services:
       - AUTH=http://moonshot-dev.mitre.org:8090/auth
       - REALM=backend_app
       - BASE_URL=http://localhost:3000
+      - DATA_TRUST_SERVICE=http://localhost:3005
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "description": "MedMorph Backend Service Client",
   "scripts": {
     "start": "npm-run-all --parallel startFrontend startBackend",
+    "startProd": "WATCH=false npm-run-all -s startFrontend startBackend",
     "startBackend": "nodemon ./bin/www",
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "description": "MedMorph Backend Service Client",
   "scripts": {
     "start": "npm-run-all --parallel startFrontend startBackend",
+    "startProd": "WATCH=false npm-run-all -s startFrontend startBackend",
```
KeeyanGhoreshi, line 9: Even though `start` works just fine, the frontend takes a bit to build and the backend starts up before it, so you get a 404 when you try to load the frontend page as it tries to find a `bundle.js` file that doesn't exist yet.

I don't think having webpack watch the files, or nodemon watch the backend files, really makes much of a difference, but I thought it would be nice to have them run sequentially (-s) instead of in parallel.

Normally it runs in parallel because the `watch` part of webpack causes the job to persist and you'd never get to the `startBackend` in sequence.  But if we don't care about watching the react files, we can just have webpack build it out and not watch, which means we can sequentially run the two.

In short, I'm looking for opinions on what we should do here.  I don't think it really matters, but perhaps it has some implications on load time I'm not thinking about.  Our options are to run it like this - building out the frontend bundle and then running the express server - or to just do a simple `npm start` so the docker container runs them in parallel, and watches the files even though they won't change.

dehall: Can we have webpack build the "production build" like `npm build` traditionally does? (Is there any difference between that and what we're doing now?)

KeeyanGhoreshi: the only difference is the `watch` property.  Traditionally `npm build` would make a build folder with `bundle.js`, which webpack does every time regardless of whether its production or development. If the `watch` property is set to `false` it is almost identical, it builds out the react files.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: "3"
+
+services:
+  medmorph_backend:
+    image: medmorph_backend
+    container_name: medmorph_backend
+    restart: on-failure
+    ports:
+      - "3000:3000"
+    environment:
```
blangley28, line 10: `DATA_TRUST_SERVICE` should also be included 
https://github.com/mcode/medmorph-backend#configuration-requirements

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:14
+
+# Create app directory
+WORKDIR /usr/src/app
```
blangley28, line 4: What are your thoughts on having a set DB to be loaded instead of whatever already exists locally? For example, this could include just the [example servers](https://github.com/mcode/medmorph-backend/wiki/Collections#examples).

blangley28 (commented):
Had a couple warnings when building the container but it still worked and I was able to run it. Can you update the "Running the Client" section of the readme to include a couple lines about using `./build-docker-image.sh` and `docker-compose up` as well as `npm start`

Dockerfile
KeeyanGhoreshi: For development purposes, I suppose it could be useful to have some set of example data loaded into the database.  It's in-memory so we could write a JS file to just set it up and load data in, right?

blangley28: I was even just thinking a version of `medmorph.db` which the Dockerfile would copy over into `/usr/src/app/medmorph.db`

KeeyanGhoreshi: ya that would also work, i don't see why not.

dehall: This is partly why I wanted an easy to run script to add servers to the DB, so that we don't have to include a copy of the DB in the repo. Otherwise we'll have to remember to remove it when we want to deploy this somewhere

KeeyanGhoreshi: that's also true, we could just run it to add a couple dummy servers in the dockerfile.

blangley28 (approved):

dehall (approved):

PR #23, opened 2021-03-03
Entire Workflow Fixes

blangley28:
Just a few minor tweaks to make the entire workflow work as expected.


src/utils/reporting_workflow.js
```diff
@@ -265,6 +266,7 @@ async function executeWorkflow(context) {
       context.currentActionSequenceStep++;
 
       // TODO: do we even need a job scheduler?
+      debug(`Waiting ${waitTime}ms until next action`);
```
dehall, line 269: Can you add an identifier to this log? When debugging with just one thing happening it's useful but in real usage when there might be a lot going on it's impossible to know what it's referring to.

PR #24, opened 2021-03-09
add logs to db

KeeyanGhoreshi:
Adds a middleware for requests, puts resources into collections based on resourceType (this was how it was being done already), and puts debug messages into the database.  The database log just adds on a timestamp and location to the message, and the files define their own location when they import the debug function.

### DIFF
--- a/src/services/messaging.js
+++ b/src/services/messaging.js
@@ -51,6 +51,7 @@ function processMessage(req, res) {
 
 function sendOperationOutcome(res, code, msg) {
   const operationOutcome = generateOperationOutcome(code, msg);
+  debug(`Message rejected with code ${code} and message ${msg}`);
   res
     .status(StatusCodes.BAD_REQUEST)
     .header('Content-Type', 'application/json')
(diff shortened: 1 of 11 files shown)


src/storage/logs.js
```diff
@@ -0,0 +1,44 @@
...
+      timestamp: Date.now(),
+      error: error,
+      location: location
+    };
+    db.insert(ERRORS, log);
+  };
+}
+
+function storeRequest(request) {
+  const log = {
```
blangley28, line 35: If there was a bearer token associated with it can we store the clientId (or just 'admin' if using the admin token)

src/services/messaging.js
```diff
@@ -37,6 +39,9 @@ function processMessage(req, res) {
     return;
   }
 
+  db.upsert(MESSAGES, messageBundle, r => r.id === messageBundle.id);
+  debug(`Message bundle ${messageBundle.id} added to database`);
+
   forwardMessageResponse(messageBundle).then(() =>
     debug(`Response to /Bundle/${messageBundle.id} forwarded to EHR`)
   );
```
blangley28, line 47: It might be a good idea to add logging to `sendOperationOutcome` function below so whenever one is returned we have a log. This could either be a log statement or even storing the OperationOutcome resource in the collection.

blangley28 (changes requested):
Looks good. Haven't checked logging was missed anywhere but this certainly seems like it hits everything.

blangley28 (approved):

danlee1025 (approved):
LGTM

PR #26, opened 2021-03-10
Added generic CRUD controller

danlee1025:
- Added `crudHandler` to generate CRUD endpoints for collections.
- Removed `serverRouter` and `/servers` endpoint since it is being created by the generic controller.
    - If the `serverRouter` was doing anything special let me know and I'll re-add it. I took a quick glance and it didn't look like it was doing anything additional.


dehall (approved):

src/app.js
```diff
@@ -45,13 +46,18 @@ passport.use(
 // Open Routes
 app.use('/.well-known', wellKnownRouter);
 app.use('/public', publicRouter);
+
+// Routes for collections
+Object.values(collections).forEach(collectionName => {
+  app.use(`/${collectionName}`, genericController(collectionName));
```
blangley28, line 52: This misses out on the authorization piece
```suggestion
  app.use(`/${collectionName}`, backendAuthorization, genericController(collectionName));
```

src/handlers/crudHandler.js
```diff
@@ -0,0 +1,65 @@
...
+function getByIdHandler(collectionName, req, res) {
+  const { id } = req.params;
+  const result = db.select(collectionName, r => r.id === id);
+  result[0] ? res.send(result[0]) : res.sendStatus(StatusCodes.NOT_FOUND);
+}
+
+function updateHandler(collectionName, req, res) {
+  const { id } = req.params;
+  const changedItem = req.body;
+  db.update(
```
blangley28, line 24: Thoughts on using upsert here?

blangley28 (changes requested):
Works as expected, two minor comments. Also the src/routes/server.js file can probably be deleted now

src/app.js
dehall: I told him leave it open for easier testing since backendAuthorization isn't the right auth for this anyway - that one is for other apps to connect to this via SMART backend services auth. 
For these we're going to want to use an "admin user" auth so that logged in users can access these, so hopefully we get one of those for free with your upcoming login changes. (And if you don't need one for that task we'll make it a new task)

danlee1025: Authorization will be added in the future so I'll just add a comment so we don't forget.

src/handlers/crudHandler.js
danlee1025: Done!

blangley28 (approved):

PR #27, opened 2021-03-10
resolve js files

KeeyanGhoreshi:
Webpack config was missing a line, missed it when it was originally pulled in because of the way things were set up.  There were no errors with `App.jsx`, I think this is because `index.jsx` is the entry point and imports it, so it wasn't caught in the original pull request.

This should allow importing of js files.  Right now webpack isn't considering js files in the path valid modules to load, this will solve that problem.  If you want to test, just copy the missing line into your own webpack config file if its experiencing errors or checkout this branch and create some example directories and `.jsx` files to import.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,6 +7,7 @@ module.exports = {
 		filename: 'bundle.js'
 	},
 	watch: watch,
+    resolve: { extensions: ["*", ".js", ".jsx"] },
 	module: {
 		rules: [
 			{
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,7 +7,7 @@ module.exports = {
 		filename: 'bundle.js'
 	},
 	watch: watch,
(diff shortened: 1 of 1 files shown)


dehall:
So just to clarify, looking at the docs I think the default, if not specified, is `['.wasm', '.mjs', '.js', '.json']`, so it wasn't considering our jsx files but if our files had .js extensions they would be fine. Not clear what the * entry does though. Check for any file extension? Does that make specifying js/jsx unnecessary?

https://webpack.js.org/configuration/resolve/#resolveextensions

KeeyanGhoreshi:
The example I was looking at for webpack config setup included a wildcard in their extensions, but I'm also not sure why.  App runs fine without it, and webpack doesn't seem to include it in their own example.  I'll just remove it for now.

blangley28 (approved):

dehall (approved):

PR #28, opened 2021-03-12
UI auth

blangley28:
Add authentication to the admin UI. Implements a very basic UI to login/logout. Added authentication to collection route.


src/ui/components/Auth/Login.jsx
```diff
@@ -0,0 +1,55 @@
...
+          {
+            email: emailTextField.current.value,
+            password: passwordTextField.current.value
+          },
+          { withCredentials: true }
+        )
+        .then(() => {
+          cache.invalidateQueries('authorized-user');
+        })
+        .catch(err => console.log(err));
```
KeeyanGhoreshi, line 26: Not sure how much we care about the UI, but this could display on the UI if the authorization fails.  Feels weird clicking "login" and having nothing happen.

src/ui/components/Auth/Login.jsx
```diff
@@ -0,0 +1,55 @@
+import './auth.css';
+import React, { memo, useCallback, useRef } from 'react';
+import { TextField, Button } from '@material-ui/core';
+import axios from 'axios';
+import { useQueryCache } from 'react-query';
+
+const Login = () => {
+  const cache = useQueryCache();
```
KeeyanGhoreshi, line 8: I tried looking up what this hook does and the docs for it seem to say it's getting deprecated as well, though it looks like it's only really changing names so it shouldn't be a problem to change.

KeeyanGhoreshi (changes requested):
express-session has some deprecation warnings, might want to change that if we can

src/app.js
```diff
@@ -43,14 +54,30 @@ passport.use(
   )
 );
 
+// Set up passport to use local config
+passport.use(new LocalStrategy({ usernameField: 'email' }, localConfig));
+passport.serializeUser((user, done) => {
+  done(null, user.uid);
+});
+passport.deserializeUser((uid, done) => {
+  // In the future, we might store user info (name, roles, etc) in Mongo via the passport.authenticate cb.
```
dehall, line 63: super minor, but the reference to Mongo doesn't really make sense since we aren't using mongo. plus the line length warning

dehall (approved):

src/app.js
blangley28: Removed mongo but left the comment

src/ui/components/Auth/Login.jsx
blangley28: Updated to use QueryClient

src/ui/components/Auth/Login.jsx
blangley28: Added a snackbar alert. I kept the console.error. Let me know if you think this should be deleted

blangley28:
> express-session has some deprecation warnings

Fixed the warnings by setting resave and saveUninitialized values

KeeyanGhoreshi (approved):

PR #29, opened 2021-03-15
Use fullUrl to access resources in DB

blangley28:
Replace all uses of insert/upsert/select for FHIR resources which use id to instead use fullUrl. This required a little refactoring, most notably changing subscription notifications to include the fullUrl. I did not update non-fhir collections such as servers.
I did some testing but since this touches a lot of different pieces looking for some help to make sure I didn't accidentally break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@material-ui/lab": "^4.0.0-alpha.57",
     "axios": "^0.21.1",
     "babel-eslint": "^10.1.0",
+    "base64url": "^3.0.1",
     "debug": "~2.6.9",
     "dotenv": "^8.2.0",
     "eslint": "^7.19.0",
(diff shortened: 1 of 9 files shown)


KeeyanGhoreshi (commented):
Looks good.  Will these changes affect the generic crud handler in any way?  Wouldn't the URL have to be base64 decoded?

blangley28:
> Will these changes affect the generic crud handler in any way? Wouldn't the URL have to be base64 decoded?

I didn't touch the crud handler at all. I suppose I should add a fullUrl route too though

blangley28:
Updated so the GET / handler now uses query parameters for `id` and `fullUrl`. Also moved the route to `/collection/{collectionname}`

KeeyanGhoreshi (approved):

dehall (approved):

PR #30, opened 2021-03-16
Display collections

danlee1025:
- Added page to display collections
    - Are there specific headers we wanted for each collection?

I still need to add some styling and prop validation but the basic functionality for displaying the db collections is ready for review.


src/ui/components/Collections.jsx
```diff
@@ -0,0 +1,85 @@
...
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow
+} from '@material-ui/core';
+import ResourceCell from './ResourceCell';
+
+const Collections = () => {
+  const [data, setData] = useState({ collectionName: '', headers: [], data: [] });
+  const collections = [
```
blangley28, line 16: Is there a way we can expose this on the backend. That way instead of updating multiple places simply modifying `src/storage/collections.js` would update everything

src/ui/components/Collections.jsx
```diff
@@ -0,0 +1,85 @@
...
+    axios.get(`http://localhost:3000/${selectedOption.value}`).then(response => {
+      setData({
+        collectionName: selectedOption.value,
+        headers: getHeaders(selectedOption.value),
+        data: response.data
+      });
+    });
+  };
+
+  const getHeaders = collectionName => {
```
blangley28, line 34: FHIR resource collections should have `fullUrl` too since that is now their "key"

src/ui/components/Collections.jsx
```diff
@@ -0,0 +1,85 @@
...
+  const collections = [
+    { value: 'servers', label: 'Servers' },
+    { value: 'endpoints', label: 'Endpoints' },
+    { value: 'subscriptions', label: 'Subscriptions' },
+    { value: 'plandefinitions', label: 'PlanDefinitions' },
+    { value: 'logs', label: 'Logs' }
+  ];
+
+  const onSelect = selectedOption => {
+    axios.get(`http://localhost:3000/${selectedOption.value}`).then(response => {
```
blangley28, line 25: Consider using react-query `useQuery` here. It won't really change anything now but it will allow us to easily update the display when the data changes.

src/ui/components/ResourceCell.jsx
```diff
@@ -0,0 +1,23 @@
...
+  const [open, setOpen] = useState(false);
+
+  return (
+    <div>
+      {!open && <Button onClick={() => setOpen(!open)}> open </Button>}
+      {open && (
+        <div>
+          <Button onClick={() => setOpen(!open)}> close</Button>
+          <TableCell key={cellKey}>
+            <pre>{JSON.stringify(resource, null, 2)}</pre>
```
blangley28, line 14: It would be really cool if you used something like https://www.npmjs.com/package/react-json-view here (this was just a quick google search but it seems easy enough to implement)

blangley28 (changes requested):

src/ui/components/Collections.jsx
danlee1025: yea I was thinking about this but we need to manually add in headers for each collection unless you can think of some generic headers for all the collections.

blangley28: Hmm, yea without refactoring the database collections to be something like {fullUrl: 'blah', resource: {}} it would be hard... I honestly think we should do that but probably not here

src/ui/components/Collections.jsx
```diff
@@ -0,0 +1,85 @@
...
+import ResourceCell from './ResourceCell';
+
+const Collections = () => {
+  const [data, setData] = useState({ collectionName: '', headers: [], data: [] });
+  const collections = [
+    { value: 'servers', label: 'Servers' },
+    { value: 'endpoints', label: 'Endpoints' },
+    { value: 'subscriptions', label: 'Subscriptions' },
+    { value: 'plandefinitions', label: 'PlanDefinitions' },
+    { value: 'logs', label: 'Logs' }
```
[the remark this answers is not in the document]
danlee1025: Not sure, for now I added whatever is in `collections.js`... Won't be too hard to add if we decide we want to view them.

src/ui/components/Collections.jsx
danlee1025: Updated to use this!

KeeyanGhoreshi (approved):

blangley28 (approved):
LGTM. Will merge once resource-fullurl has been approved.

PR #31, opened 2021-03-23
Dashboard layout

KeeyanGhoreshi:
This creates the dashboard layout according to the designs.

I couldn't quite get the font right.  After you log in you will be brought to the dashboard.  I've relocated the collections display into the various collection tabs.  The icon on the top right will open up the menu, which lets you logout.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,15395 @@
 {
   "name": "medmorph-backend",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.0.1",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "@material-ui/core": "^4.11.3",
+        "@material-ui/icons": "^4.11.2",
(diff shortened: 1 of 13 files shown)


blangley28 (commented):
This looks really good. A few comments:

- [ ] Can you include wdyr for development. I want to make sure we stay on top of that from the beginning. I notice anonymous functions being passed into props and if its not a material ui component that will cause a warning from wdyr
- [x] When hovering over the icons in the top right the cursor should be a pointer so users know they can click on them
- [x] I think there should be a little more top margin the right drawer. The settings button is quite close to the top
- [ ] I'm also wondering if clicking "Medmorph Admin Console" should bring the user back to the Dashboard tab

dehall:
This might just be Firefox but I see a white bar on the side of the sidebar 
image

At smaller sizes, I think it's the scrollbar? 
image

Anything we can do about that?

KeeyanGhoreshi:
I can't seem to reproduce the white bar.  The scrollbar is correctly colored for me on firefox.

dehall (approved):

blangley28 (approved):

PR #32, opened 2021-03-29
Receive R5 Subscriptions

blangley28:
Update notification endpoints to accept R5 Backport Subscriptions which come as a Bundle. Still unclear as to whether or not the PUT `{base url}/{resourceType}/{resourceId}` is used in the backport subscriptions

### DIFF
--- a/src/routes/subscriptions.js
+++ b/src/routes/subscriptions.js
@@ -1,128 +1,31 @@
-const { StatusCodes } = require('http-status-codes');
-
 const express = require('express');
 const router = express.Router();
-const axios = require('axios');
-const db = require('../storage/DataAccess');
-const { startReportingWorkflow } = require('../utils/reporting_workflow');
-const { getServerById } = require('../storage/servers');
-const {
-  refreshKnowledgeArtifact,
-  getEndpointId,
-  subscriptionsFromPlanDef,
-  postSubscriptionsToEHR,
-  getBaseUrlFromFullUrl
-} = require('../utils/fhir');
-const { getAccessToken } = require('../utils/client');
-const { PLANDEFINITIONS, ENDPOINTS } = require('../storage/collections');
-const { default: base64url } = require('base64url');
-const debug = require('../storage/logs').debug('medmorph-backend:subscriptions');
+const notificationService = require('../services/subscriptions');
 
 /**
  * Notification from KA Repo to refresh artifacts. The id param is the
  * server id from the database
  */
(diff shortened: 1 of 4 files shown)


dehall (approved):

src/routes/subscriptions.js
```diff
@@ -1,128 +1,31 @@
...
-      debug(`Fetched ${endpointFullUrl}`);
-    }
-  });
-
-  const subscriptions = subscriptionsFromPlanDef(planDefinition, server.endpoint);
-  postSubscriptionsToEHR(subscriptions);
-
-  res.sendStatus(StatusCodes.OK);
-});
+router.put('/ka/:id/:resource/:resourceId', notificationService.knowledgeArtifact);
```
danlee1025, line 16: Do we still need this endpoint? It doesn't look like `knowledgeArtifact` function uses the `resourceId` param.

src/routes/subscriptions.js
```diff
@@ -1,128 +1,31 @@
...
- * @param {string} fullUrl - the base 64 url encoded fullUrl of the PlanDefinition
- * @returns the PlanDefinition with the given id or null if not found
- */
-function getPlanDef(fullUrl) {
-  const decodedFullUrl = base64url.decode(fullUrl);
-  const resultList = db.select(PLANDEFINITIONS, s => s.fullUrl === decodedFullUrl);
-  if (resultList[0]) return resultList[0];
-  else return null;
-}
+router.put('/:fullUrl/:resource/:resourceId', notificationService.reportTrigger);
```
danlee1025, line 29: Same as the other endpoint, I don't see `resourceId` being used in `reportTrigger`

danlee1025 (changes requested):

src/routes/subscriptions.js
blangley28: So `resource` and `resourceId` are not used but I left this endpoint in... it's unclear at this point whether the backport will POST to {notification url} and/or PUT to {notification url}/Bundle/{bundleid}. So I decided to keep both endpoints for now. However changing this to `router.put('ka/:id, notificationService.knowledgeArtifact);` will give a 404 error for PUT to {notification url}/Bundle/{bundleid}. So even though the value isn't used its still needed for the endpoint

danlee1025: Oh ok that makes sense, thanks.

danlee1025 (approved):

PR #33, opened 2021-03-30
add table ui

KeeyanGhoreshi:
The collections now are displayed in a table as shown in the designs.  The pagination works and you can sort the columns, but none of the buttons are functional.  The only differences from the design are some minor color/font and spacing issues and the "Add server" button being replaced with a generic "Add new" button.  This button could be changed to use the actual collections name, but we'd have to do a bit of extra processing.  

Different collections can be given different parameters as well, the "Add new" button won't show up for "Logs" and "Requests".

### DIFF
--- a/src/ui/components/Collections/Collections.jsx
+++ b/src/ui/components/Collections/Collections.jsx
@@ -142,7 +142,7 @@ const Collections = props => {
               return (
                 <TableCell key={cellKey} style={{ whiteSpace: 'nowrap' }}>
                   {' '}
-                  {header === 'RESOURCE' ? (
+                  {header === 'resource' ? (
                     <ReactJson src={d} collapsed={true} enableClipboard={false} />
                   ) : (
                     d[header.value || header.toLowerCase()]
--- a/src/ui/components/Collections/Collections.jsx
+++ b/src/ui/components/Collections/Collections.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import axios from 'axios';
 import { useQuery } from 'react-query';
 import {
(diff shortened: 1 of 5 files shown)


src/ui/components/Collections/Collections.jsx
```diff
@@ -1,81 +1,215 @@
...
+    return 0;
+  }
+
+  function serverStatus(server) {
+    return server && true;
+  }
+
+  function getComparator(order, orderBy) {
+    return order === 'desc'
+      ? (a, b) => descendingComparator(a, b, orderBy.toLowerCase())
```
dehall, line 49: This might be a non-issue, but I'm concerned about doing a `toLowerCase` here -- is it possible we'll have fields with camelCase names? To be extra defensive I'd suggest we use the "official" name in the definition below and then just render it with all-caps.

src/ui/components/Collections/Collections.jsx
```diff
@@ -1,81 +1,215 @@
...
+    setOrderBy(property);
+  };
+
+  const handleChangeRowsPerPage = event => {
+    setRowsPerPage(parseInt(event.target.value, 10));
+    setPage(0);
+  };
 
   const { data } = useQuery(['collections', { selectedCollection }], () =>
     axios.get(`http://localhost:3000/collection/${selectedCollection}`)
```
dehall, line 76: I realize this is not your change but I'm just seeing it now. This should not be hardcoded to localhost:3000. Will this work?
```suggestion
    axios.get(`/collection/${selectedCollection}`)
```

dehall (commented):
Looks really nice. One thing stands out though - the table seems to be slightly larger than its container and it can scroll vertically within a small space:
image
image
image

KeeyanGhoreshi:
The table is the right size for me when everything is one line, like in your example with the logs.  It does resort to scrolling if you open a resource or change the rows per page.  Do we want it to instead extend vertically down the page?

dehall:
Yeah that would be ideal - have it extend vertically down the page. Hopefully that's an easy change

src/ui/components/Collections/Collections.jsx
```diff
@@ -1,81 +1,215 @@
...
 
 const Collections = props => {
+  const classes = useStyles();
   const { selectedCollection } = props;
+  const [page, setPage] = useState(0);
+  const [rowsPerPage, setRowsPerPage] = useState(10);
+  const [order, setOrder] = useState('asc');
+  const [orderBy, setOrderBy] = useState('');
+
+  const handleChangePage = (event, newPage) => {
```
blangley28, line 29: The functions `handleChanhePage`, `handleRequestSort`, and `handleChangeRowsPerPage` should all use `useCallback` since they are passed into a react component.

src/ui/components/Collections/SortedTableHead.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { TableCell, TableHead, TableRow, TableSortLabel } from '@material-ui/core';
+import PropTypes from 'prop-types';
+
+function SortedTableHead(props) {
+  const { classes, order, orderBy, onRequestSort, headers } = props;
+  const createSortHandler = property => event => {
```
blangley28, line 7: This should be a callback too

src/ui/components/Collections/styles.jsx
```diff
@@ -0,0 +1,80 @@
...
+    addButton: {
+      float: 'right',
+      marginRight: '25px',
+      marginTop: '17px',
+      color: theme.palette.common.white
+    },
+    break: {
+      width: '100%',
+      height: '0',
+      borderTop: '1px solid #eaeaeb'
```
blangley28, line 13: Should this color be added to the palette?

blangley28 (changes requested):
Looks great. Agree with what Dylan says above. Noticed this small bug when I was demoing it yesterday. Steps to reproduce:

1. Click on a collection with many pages (such as Logs)
2. Page through to the 2nd or 3rd page
3. Click on a collection with only a couple items (such as Subscriptions in my case or Servers)
4. Notice the pagination now says "51-2 of 2" instead of jumping back to the first page and showing "1-2 of 2".

Screen Shot 2021-04-01 at 9 06 24 AM

src/ui/components/Collections/Collections.jsx
```diff
@@ -1,81 +1,219 @@
...
+      .slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage)
+      .map((d, i) => {
+        return (
+          <TableRow key={`${selectedCollection}-${i}`} classes={{ root: classes.tableRow }}>
+            {infoBundle.headers.map((header, j) => {
+              const cellKey = `${i}-${j}`;
+              return (
+                <TableCell key={cellKey} style={{ whiteSpace: 'nowrap' }}>
+                  {' '}
+                  {header === 'RESOURCE' ? (
```
dehall, line 145:
```suggestion
                  {header === 'resource' ? (
```

dehall (approved):

blangley28:
On small screens there is horizontal overflow. Perhaps set `overflow-x: scroll`?

blangley28 (approved):
There's nothing wrong with how it works now but I do feel like the pagination should be fixed (and not scroll with the table). I'll approve now though and if you think its fine go ahead and merge it

PR #35, opened 2021-04-08
Manually trigger report from the UI

blangley28:
Update the backend and the UI to support manually triggering reports. A couple comments:
1. Right now we can trigger any report, I'm not sure if we want to check specifically for the `manual-notification` named-event. 
2. I think the snackbar alert looks a little weird on smaller screens since it centers to the screen and not the dashboard component. Reached out to Noranda for feedback.

For testing I have been using the following Encounter 
```
{
  "resourceType": "Encounter",
  "id": "encounter-manual-trigger-example",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Encounter with patient pat01</div>"
  },
  "status": "in-progress",
  "class": {
    "system": "http://terminology.hl7.org/CodeSystem/v3-ActCode",
    "code": "IMP",
    "display": "inpatient encounter"
  },
  "subject": {
    "reference": "Patient/pat01"
  }
}
```


src/ui/components/Dashboard/Dashboard.jsx
```diff
@@ -1,61 +1,151 @@
...
+        axios
+          .post('/index/trigger', { planDefFullUrl, resource }, { withCredentials: true })
+          .then(() => {
+            setIsLoading(false);
+            setFileName('');
+            setPlanDefFullUrl('');
+            setAlert('Manual Report Triggered Successfully!', 'success');
+          })
+          .catch(err => {
+            console.error(err);
```
dehall, line 56: I think we want to `setIsLoading(false)` here as well, so if there's an error it unlocks the button and the user can try again.

src/ui/components/Dashboard/Dashboard.jsx
```diff
@@ -1,61 +1,151 @@
...
+    [setPlanDefFullUrl]
+  );
+
+  const handleChooseFile = useCallback(() => {
+    if (importFileRef?.current?.files?.[0]) setFileName(importFileRef.current.files[0].name);
+  }, [importFileRef, setFileName]);
+
+  const handleTrigger = useCallback(
+    event => {
+      event.preventDefault();
```
dehall, line 40: One thing I noticed was that you can click the trigger button without choosing a file or a PlanDef, and it just hung. We don't need to get all fancy with disabling buttons or any of that, but could there be a simple `if` check here to make sure both the plandef and patient/encounter are set before making the request, and if not post an alert saying something like "PlanDefinition and Patient/Encounter are required"?

src/ui/components/Dashboard/Dashboard.jsx
```diff
@@ -1,61 +1,151 @@
...
-    <div className={classes.container}>
-      <NavBar />
-      <Sidebar tabs={tabs} callback={setContentKey} selected={contentKey} />
-      <main className={classes.content}>
-        <div className={classes.spacer} />
-        {content ? content.component : 'Default'}
-      </main>
+    <div className={classes.dashboardContainer}>
+      <Paper elevation={3} className={classes.dashboardCard}>
+        <p>Manually Trigger Notification</p>
```
dehall, line 81: Can the Manually Trigger Notification be a separate tab instead of on the main dashboard? I'm open to other opinions but it doesn't feel like it's important enough to be on the dashboard. (I do realize we don't have anything else on the dashboard)

dehall (commented):
Not sure if this really matters, but I noticed from the network tab (which I only coincidentally had open) that the `/PlanDefinition` and `/active` endpoints are hit every time the page is reactivated. Not just refreshed, but if you switch from one tab to another and back, it hits those endpoints again. Might be worth a quick investigation on why it's doing that, assuming it's not intentional

src/ui/components/Dashboard/Dashboard.jsx
blangley28: I'm open to moving it. It doesn't take up a ton of real estate though so I feel like having its own tab is overkill. Perhaps if we other "quick actions" those could all be a new tab?

dehall (approved):

src/utils/auth.js
```diff
@@ -70,9 +70,23 @@ function userAuthorization(req, res, next) {
...
+  else if (adminToken && token === adminToken) return next();
+  else return passport.authenticate('keycloak', { session: false })(req, res, next);
+}
+
 module.exports = {
   generateToken,
   backendAuthorization,
   subscriptionAuthorization,
-  userAuthorization
+  userAuthorization,
```
danlee1025, line 90: Is anything still using `userAuthorization`? We can probably get rid of this function and the export if it's not being used.

danlee1025 (commented):
Looks good! Just comment to remove an unused function, will approve once that's addressed and this branch is rebased.

danlee1025 (approved):

PR #36, opened 2021-04-09
Add AWS Docker Compose

blangley28:
Add a docker compose file to run on AWS

### DIFF
--- a/docker/aws-docker-compose.yml
+++ b/docker/aws-docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3"
+
+services:
+  medmorph_backend:
+    image: medmorph_backend
+    container_name: medmorph_backend
+    restart: on-failure
+    ports:
+      - "3000:3000"
+    environment:
+      - AUTH=http://ec2-3-211-58-180.compute-1.amazonaws.com:8090/auth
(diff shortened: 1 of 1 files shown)


docker/aws-docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+
+services:
+  medmorph_backend:
+    image: medmorph_backend
+    container_name: medmorph_backend
+    restart: on-failure
+    ports:
+      - "3000:3000"
+    environment:
+      - AUTH=http://ec2-3-211-58-180.compute-1.amazonaws.com:8090/auth
```
dehall, line 11: Since these are all running on the same system, and via http not https, can we just use `localhost` for these urls?

blangley28: So that should work but went to go test it I ran into a bigger issue - the only place this is being used is for the $process-message operation asynchronous response. But the library doesn't even add an authorization header so I'll create a task to fix that

blangley28: Update: this does not work. I generated a client_assertion myself to get an access token I get unauthorized. It must be something internal to passport's keycloak strategy where it is checking ISS/AUD.

PR #37, opened 2021-04-12
Remove keycloak, switch to allowing any JWT

dehall:
Removes the specific dependency on keycloak, and allows the use of any JWT source for authorization.

Heavily inspired by this example: https://github.com/auth0/node-jwks-rsa/tree/master/examples/passport-demo


blangley28 (approved):
One super minor comment. Looks good. Tested locally and was able to authorize correctly.

danlee1025 (approved):
LGTM

PR #38, opened 2021-04-12
modify rows

KeeyanGhoreshi:
The main functionality is implemented, you can click on the edit/delete/add buttons to do the intended action.  The only part that is a bit uncertain is the "add new" part.  Right now the UI is just an extra, special row in the table that's fixed to the top which is allowed to be edited like any other row.  This means if a certain field is not editable in an existing entry, it is also not editable when creating a new entry.

There's also the question of "editing" the resource associated with plandefinitions/endpoints/subscriptions.  Right now they're uneditable, but we could add a textbox for pasting JSON in, or maybe a file selector.

### DIFF
--- a/src/ui/components/Collections/CollectionRow.jsx
+++ b/src/ui/components/Collections/CollectionRow.jsx
@@ -30,24 +30,24 @@ function CollectionRow(props) {
     });
 
     axios.put(`/collection/${selectedCollection}?id=${bundle.id}`, bundle).then(() => {
-        queryClient.invalidateQueries(['collections', {selectedCollection}]);
-        setEdit(false);
-        if(callback) {
-            callback();
-        }
+      queryClient.invalidateQueries(['collections', { selectedCollection }]);
+      setEdit(false);
+      if (callback) {
+        callback();
+      }
     });
   };
 
   const cancelSave = () => {
     setEdit(false);
-    if(callback) {
-        callback();
+    if (callback) {
+      callback();
     }
   };
 
   const deleteData = () => {
     axios.delete(`/collection/${selectedCollection}?id=${state.id}`).then(() => {
-      queryClient.invalidateQueries(['collections', {selectedCollection}]);
+      queryClient.invalidateQueries(['collections', { selectedCollection }]);
       setEdit(false);
     });
   };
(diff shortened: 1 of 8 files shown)


src/ui/components/Collections/CollectionRow.jsx
```diff
@@ -0,0 +1,145 @@
...
+  const updateData = () => {
+    const bundle = {};
+    headers.forEach(header => {
+      if (!header.viewOnly) {
+        bundle[header.value] = state[header.value];
+      }
+    });
+
+    axios.put(`/collection/${selectedCollection}?id=${bundle.id}`, bundle).then(() => {
+      queryClient.invalidateQueries('collections');
```
blangley28, line 33: Query keys can be more complex than a string. Instead of invalidating all collections queries it looks like the query key is set to `['collections', { selectedCollection }]` (Collections.jsx:83). Use the full key for all of these invalidateQueries so only the cache for that specific collection is invalidated.

blangley28 (commented):
Looking nice a few things I noticed off the bat:
1. The textboxes look a little friendlier in the designs. Might just be the border but I would ask Noranda for styling
2. When you click on "Add New" and then switch the collection it maintains the new row on the top. This should be cleared when switching tabs

dehall:
Overall looks good. We may unfortunately have to go deeper into metadata to say certain fields are required, certain fields are certain types, etc. Right now it allows users to save rows where all the fields are blank. (well except the ID is autogenerated)

Can you see if there are any libraries we could easily pull in for editing JSON? If not I'd suggest just making it a plain textarea and we can make it nicer later

KeeyanGhoreshi:
I looked at a couple libraries for JSON editing, a couple of them seemed promising but ultimately looked like they would be a bit of work to implement, and were ill-placed in the UI.  Right now its just a textarea, and you can only really paste JSON in it, but its functional.

blangley28:
Is there a reason why you can only paste JSON in it and not type JSON? Perhaps not for this task but I think it is worth investigating more.
Also the JSON doesn't seem to be editable in all collections (i.e. Add New Subscription)

KeeyanGhoreshi:
The textarea value needs to be a string, but the state needs a JSON object, so if you try to input an invalid JSON object, it will fail.  

Some aren't editable just because the 'editable' flag is set to false in the headers.

KeeyanGhoreshi:
latest edit uses a library, it allows editing of the json more seamlessly but it doesn't fit into the UI all that well, and probably needs to be brought out as a popup.

src/ui/components/Collections/Collections.jsx
```diff
@@ -86,44 +91,73 @@ const Collections = props => {
       case '':
         return [];
       case 'servers':
-        headers = [{ value: 'icon', label: '' }, 'name', 'id', 'endpoint', 'type'];
+        headers = [
+          { value: 'icon', label: '', edit: false, viewOnly: true },
+          { value: 'name', label: 'name', edit: true },
+          { value: 'id', label: 'id', edit: false },
+          { value: 'endpoint', label: 'endpoint', edit: true },
+          { value: 'type', label: 'type', edit: false }
```
blangley28, line 99: Is there a reason this is false? When adding a new one there is no way to set this

src/ui/components/Collections/styles.jsx
```diff
@@ -63,6 +63,9 @@ export default makeStyles(
         backgroundColor: theme.palette.common.grayHighlight
       }
     },
+    tableRowEdit: {
+      backgroundColor: '#e6f7f5'
```
blangley28, line 67: I'm not sure whether these two new colors will be reused but consider adding them to the palette

blangley28 (changes requested):
FYI I am still seeing all of the resources being an empty object. One other thing (and I know it wasn't this PR but its small enough to change), I get the following warning ```express deprecated res.send(status): Use res.sendStatus(status) instead src/handlers/crudHandler.js:55:7```

src/ui/components/Collections/Collections.jsx
KeeyanGhoreshi: The designs show the type as uneditable, but I guess when adding a new one you do need to be able to add it.  I'll have to figure out how to resolve that.

dehall (approved):
Looks good to me. No other comments beyond what Ben already noted

blangley28 (approved):

PR #39, opened 2021-04-21
Check Access Token Scope

danlee1025:
- Check access token scope before authorizing request
 - Added `system/*.*` and `system/*.write` scopes in addition to `system/*.read`
     - All requests are allowed when scope includes `system/*.*`
     - GET requests are allowed when scope includes `system/*.read`
     - POST, PUT, and DELETE requests are allowed when scope includes `system/*.write`
 - To test, you must first get an access token from the keycloak server. I added another client with the client_id `test` which made scopes optional. client_secret can be found on the keycloak server.
     - You should get a 401 with an "Insufficient scope" message if you don't have the correct scope.

### DIFF
--- a/src/handlers/crudHandler.js
+++ b/src/handlers/crudHandler.js
@@ -52,7 +52,7 @@ function deleteHandler(collectionName, req, res) {
     return;
   }
 
-  res.send(StatusCodes.OK);
+  res.sendStatus(StatusCodes.OK);
 }
 
 function genericController(collectionName) {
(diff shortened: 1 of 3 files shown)


blangley28 (approved):
LGTM. Only note is that it seems like Keycloak ignores the requested scope and just always gives the same thing. So to test it I had to remove `system/*.*` from the `testapp` client. Might be worth doing a little investigating to see if that is something we can control. Other than that though, the code here works as expected

danlee1025:
I created a `test` client which removes `system/*.*` as the default scope and made them all optional. The `test` client also made it easier for me to test with postman.

dehall (approved):

PR #40, opened 2021-04-21
Notifications

KeeyanGhoreshi:
Adds the UI for the notification tab, functionality for the notification bell icon, and a new collection to the database specifically for notifications under the assumption that future notifications won't all exclusively be errors.  The notification data structure should also include the userId if it can, but I've left that out for now to simplify things.

To test the UI, you can either trigger an error (that reports using the `error()` log function) or you can just visit `/index`, since I've added some testing code to create an error when visiting the test endpoint.  This should trigger the backend UI to display the new notification, which can be dismissed by pressing on the grey checkmark.

### DIFF
--- a/src/services/test_service.js
+++ b/src/services/test_service.js
@@ -1,6 +1,6 @@
 const error = require('../storage/logs').error('medmorph-backend:test');
 module.exports = (req, res, next) => {
-  error('there was an error');
+  error('test error', 'Test Service: Test Notification');
   res.send('Howdy from test service!');
   next();
 };
--- a/src/services/test_service.js
+++ b/src/services/test_service.js
@@ -1,4 +1,6 @@
+const error = require('../storage/logs').error('medmorph-backend:test')
 module.exports = (req, res, next) => {
+  error("there was an error");
   res.send('Howdy from test service!');
   next();
 };
--- a/src/services/test_service.js
+++ b/src/services/test_service.js
@@ -1,6 +1,6 @@
-const error = require('../storage/logs').error('medmorph-backend:test')
+const error = require('../storage/logs').error('medmorph-backend:test');
 module.exports = (req, res, next) => {
-  error("there was an error");
+  error('there was an error');
   res.send('Howdy from test service!');
   next();
 };
--- a/src/storage/logs.js
+++ b/src/storage/logs.js
@@ -30,11 +30,11 @@ function error(location) {
     db.insert(ERRORS, log);
 
     const notification = {
-        id: uuidv4(),
-        timestamp: Date.now(),
-        notif: notif ? notif : error,
-        viewed: false,
-        type: 'error',
+      id: uuidv4(),
+      timestamp: Date.now(),
+      notif: notif ? notif : error,
+      viewed: false,
+      type: 'error'
     };
     db.insert(NOTIFICATIONS, notification);
   };
--- a/src/storage/logs.js
+++ b/src/storage/logs.js
@@ -1,5 +1,5 @@
 const db = require('./DataAccess');
-const { LOGS, ERRORS, REQUESTS } = require('./collections');
+const { LOGS, ERRORS, REQUESTS, NOTIFICATIONS } = require('./collections');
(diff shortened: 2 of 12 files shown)


src/ui/components/Dashboard/styles.jsx
```diff
@@ -17,6 +17,18 @@ export default makeStyles(
       fontWeight: 'bold',
       margin: '0 15px 0 15px'
     },
+    badge: {
+      backgroundColor: theme.palette.common.maroon,
+      top: '35%',
+      right: '35%',
+      border: '2px solid white',
```
blangley28, line 24:
```suggestion
      border: '2px solid theme.palette.common.white',
```

src/services/test_service.js
```diff
@@ -1,4 +1,6 @@
+const error = require('../storage/logs').error('medmorph-backend:test');
 module.exports = (req, res, next) => {
+  error('there was an error');
```
blangley28, line 3: Is the plan to delete this? I think its fine to keep it as a test but the text should reflect that 
```suggestion
  error('test error', 'Test Service: Test Notification');
```

src/ui/components/Dashboard/NavBar.jsx
```diff
@@ -8,11 +8,18 @@ import TextField from '@material-ui/core/TextField';
...
-function NavBar() {
+function NavBar(props) {
   const [open, setOpen] = useState(false);
-
+  const { newNotifs, callback } = props;
   const classes = useStyles();
+
+  const setNotif = useCallback(() => {
+    callback('notifications');
+  });
```
blangley28, line 21: Since `callback` is always used. just to set the screen can the name be updated to `setNotificationContent` or something like that?

src/ui/components/Notifications/Notifications.jsx
```diff
@@ -0,0 +1,111 @@
...
+  const updateNotif = useCallback(notif => {
+    return () => {
+      notif.viewed = true;
+      axios.put(`/collection/notifications?id=${notif.id}`, notif).then(() => {
+        queryClient.invalidateQueries('notifications');
+      });
+    };
+  });
+
+  const getElapsed = time => {
```
blangley28, line 26: In the Pathway Builder we used the [moment](https://www.npmjs.com/package/moment) library. The README says it is now a legacy project but I would use a library here instead of doing this manually

blangley28 (changes requested):

src/storage/logs.js
```diff
@@ -28,6 +28,15 @@ function error(location) {
       location: location
     };
     db.insert(ERRORS, log);
+
+    const notification = {
+      id: uuidv4(),
+      timestamp: Date.now(),
+      notif: notif ? notif : error,
```
danlee1025, line 35:
```suggestion
      notif: notif || error,
```

src/ui/components/Dashboard/DashboardLayout.jsx
```diff
@@ -1,49 +1,76 @@
...
 
 function DashboardLayout() {
   const classes = useStyles();
   const [contentKey, setContentKey] = useState('dashboard');
+
+  const { data } = useQuery(['notifications'], () => axios.get(`/collection/notifications`));
+
+  const unviewedNotifs = useMemo(() => {
+    if (data) {
+      return data.data.filter(notif => {
```
danlee1025, line 19: Since it's just one line in the function you can remove the `return` and curly braces
```suggestion
      return data.data.filter(notif => (!notif.viewed));
```

src/ui/components/Dashboard/DashboardLayout.jsx
```diff
@@ -1,49 +1,76 @@
...
+        return !notif.viewed;
+      });
+    } else {
```
danlee1025, line 22: Unnecessary `else` since something's guaranteed to be returned in the `if`

blangley28 (approved):

danlee1025 (approved):

PR #41, opened 2021-04-25
Delete resources from server

danlee1025:
Deleting a Subscription or PlanDefinition from the backend app deletes the corresponding Subscriptions on the EHR/KA server.

### DIFF
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -465,7 +465,7 @@ function getBaseUrlFromFullUrl(fullUrl) {
 
 /**
  * Returns the last part of path from url
- *  Will return id from url in the form of {baseUrl}/{resourceType}/{id}
+ * For notification endpoints this will be the base64 encoded fullUrl of the PlanDefinition
  *
  * @param {string} url
  * @returns {string} last part of path from url
@@ -487,7 +487,7 @@ function getPlanDef(fullUrl) {
 }
 
 /**
- * Deletes the resource
+ * Deletes the resource from external server
  *
  * @param {Resource} resource
  */
--- a/src/utils/fhir.js
+++ b/src/utils/fhir.js
@@ -292,12 +292,13 @@ function subscriptionsFromPlanDef(planDef, serverUrl) {
 /**
  * Returns subscriptions in local DB associated with plandef
  *
- * @param {PlanDefinition} planDef - PlanDefinition resource
+ * @param {String} fullUrl - fullUrl of PlanDefinition resource
  * @returns list of Subscription resources
  */
-function getSubscriptionsFromPlanDef(planDef) {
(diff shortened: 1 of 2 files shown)


src/utils/fhir.js
```diff
@@ -447,6 +463,17 @@ function getBaseUrlFromFullUrl(fullUrl) {
   return temp.substring(0, temp.lastIndexOf('/'));
 }
 
+/**
+ * Returns the last part of path from url
+ *  Will return id from url in the form of {baseUrl}/{resourceType}/{id}
```
blangley28, line 468: I think this comment is a little misleading. 
```suggestion
 * Returns the last part of path from url
 * For notification endpoints this will be the base64 encoded fullUrl of the PlanDefinition
```

src/utils/fhir.js
```diff
@@ -289,6 +289,22 @@ function subscriptionsFromPlanDef(planDef, serverUrl) {
   }, []);
 }
 
+/**
+ * Returns subscriptions in local DB associated with plandef
+ *
+ * @param {PlanDefinition} planDef - PlanDefinition resource
+ * @returns list of Subscription resources
+ */
+function getSubscriptionsFromPlanDef(planDef) {
```
blangley28, line 298: If you wanted this could just take in `fullUrl` instead of the entire resource

src/utils/fhir.js
danlee1025: Yea I definitely didn't know how to phrase this... I wanted to leave it generic since it can be used for other things.

src/utils/fhir.js
```diff
@@ -459,18 +487,35 @@ function getPlanDef(fullUrl) {
...
+ * Deletes the resource from external server
+ *
+ * @param {Resource} resource
+ */
+async function deleteResource(resource) {
+  const { fullUrl } = resource;
+  if (fullUrl) {
+    const token = await getAccessToken(getBaseUrlFromFullUrl(fullUrl), db);
+    const headers = { Authorization: `Bearer ${token}` };
+    axios.delete(fullUrl, { headers });
```
dehall, line 500: I need to load some data into my db to check what all the fullUrls are, but at a glance this generic approach to deleting feels a little scary. Not that it can get called with completely arbitrary data, it should only get called with subscriptions, but still.
As of now the only thing we want to delete on an external server is subscriptions, and that's only on the EHR server. I don't think we want to delete the subscription from the KAR, because that's generic on all PlanDefinitions. (if a subscription there even exists) So just to be safe and prevent any chance of badness, how would you feel about making this more explicit and hardcoded to Subscription resources on the EHR?

danlee1025: Sounds good to me, updated the function.

blangley28 (approved):

dehall (approved):

PR #42, opened 2021-04-30
Small Fixes for Running on AWS

blangley28:
Three fixes to get it working on AWS:
1. Include the clientId for servers in the UI. If this field is not included when creating the server the authorization will fail. I know we are removing auth for the connectathon but this won't break anything without auth (it will just be extra info there which means nothing).
2. Update the docker-compose files to include the DEBUG and ADMIN_TOKEN environment variables
3. Ensures errors are logged with a helpful message and id

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -102,9 +102,8 @@ const baseIgActions = {
         }
       )
       .catch(err => {
-        error(
-          `Error submitting Bundle/${context.reportingBundle.id} to ${context.client.dest}\n${err.message}`
-        );
+        const bundleId = context.reportingBundle.id;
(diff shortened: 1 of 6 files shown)


src/utils/actions.js
```diff
@@ -102,7 +102,9 @@ const baseIgActions = {
         }
       )
       .catch(err => {
-        error(err);
+        error(
+          `Error submitting Bundle/${context.reportingBundle.id} to ${context.client.dest}\n${err.message}`
```
danlee1025, line 106: Lint warning.

danlee1025 (commented):
Changes look good, just a lint warning and needs a rebase.

danlee1025 (approved):

dehall (approved):

PR #43, opened 2021-04-30
login page ui

KeeyanGhoreshi:
changes the login page to have a simple UI.  The forget password and keep my logged in options don't do anything for now.  But you can press enter to log in after putting in credentials now.

In order to get the linter to work, I had to add an extra `eslintrc` file so that the UI files would allows global browser variables like `document`.  

You can test the changes by running the app and going to the home page to log in.

### DIFF
--- a/src/ui/components/Auth/Login.jsx
+++ b/src/ui/components/Auth/Login.jsx
@@ -1,26 +1,47 @@
-import './auth.css';
-import React, { memo, useCallback, useRef, useState } from 'react';
+import React, { memo, useCallback, useState, useEffect } from 'react';
 import { TextField, Button, Snackbar } from '@material-ui/core';
 import { Alert } from '@material-ui/lab';
 import axios from 'axios';
 import { useQueryClient } from 'react-query';
+import useStyles from './styles';
 
 const Login = () => {
   const queryClient = useQueryClient();
-  const emailTextField = useRef();
-  const passwordTextField = useRef();
+  const classes = useStyles();
   const [message, setMessage] = useState(null);
-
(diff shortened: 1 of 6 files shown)


src/ui/components/Auth/Login.jsx
```diff
@@ -1,26 +1,47 @@
...
+        onSubmit();
+      }
+    };
+    document.addEventListener('keydown', listener);
+    return () => {
+      document.removeEventListener('keydown', listener);
+    };
+  }, [username, password]);
+
+  const _setUsername = event => {
```
blangley28, line 29: From my experience the underscore version is the one to not use. So the state would be defined as `[username, _setUsername] = ....` and then this would be `setUsername`. Also these two methods should be callbacks

src/ui/components/Auth/styles.jsx
```diff
@@ -0,0 +1,100 @@
+import { makeStyles } from '@material-ui/core/styles';
+export default makeStyles(
+  theme => ({
+    adminBar: {
+      height: '95px',
+      backgroundColor: '#8b72d6',
```
blangley28, line 6: Colors should be added to the palette

blangley28 (commented):
Looks really good, only thing visually which seems a little off from the designs is the font weight in a couple places.
Designs:
Screen Shot 2021-05-03 at 8 44 10 AM

Current:
Screen Shot 2021-05-03 at 8 43 47 AM

dehall (approved):
I'm a simple man - when I can hit enter to login, I click approve

blangley28 (approved):

PR #44, opened 2021-05-03
Fixes based on sandbox testing

dehall:
This PR brings in a few fixes based on testing against the Epic and Cerner sandboxes:

- Cerner doesn't support wildcard scopes, so now we can configure a custom set of scopes to use per server
- Cerner uses the `client_secret_basic` method of authentication which uses a client_secret instead of a JWT
- Epic's smart-configuration isn't at `{baseUrl}/.well-known/smart-configuration` so if getting that URL fails, we can try to get the token endpoint via the server metadata
- Posting Subscriptions to the subscription-proxy fails if we don't specify the content-type

### DIFF
--- a/src/utils/client.js
+++ b/src/utils/client.js
@@ -36,6 +36,10 @@ async function connectToServer(url) {
     }
   };
 
+  // NOTE: client_secret_basic authorization is not part of the backend auth spec
+  // (which requires the use of the signed authentication JWT)
+  // but it is included here to enable connecting to the Cerner sandbox,
+  // which currently only supports the client_secret_basic auth type
   if (secret) {
     delete props.client_assertion_type;
     delete props.client_assertion;
@@ -45,6 +49,7 @@ async function connectToServer(url) {
 
     headers.headers['Authorization'] = `Basic ${base64}`;
   }
+  // end client_secret_basic logic
 
   // Get access token from auth server
   const data = await axios
--- a/src/utils/client.js
+++ b/src/utils/client.js
@@ -86,8 +86,29 @@ async function getAccessToken(url) {
  * @returns token_endpoint
  */
 async function getTokenEndpoint(url) {
-  const response = await axios.get(`${url}/.well-known/smart-configuration`);
-  return response.data.token_endpoint;
+  try {
+    const response = await axios.get(`${url}/.well-known/smart-configuration`);
+    return response.data.token_endpoint;
+  } catch (e) {
+    try {
+      // sometimes the smart-config is in a non-standard place,
+      // so let's try the server capability statement
+      const response = await axios.get(`${url}/metadata`);
+
+      const rest = response.data.rest;
+      const serverRest = rest.find(r => r.mode === 'server');
+      const extensions = serverRest.security.extension;
+      const oauth = extensions.find(
+        e => e.url === 'http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris'
+      );
+      return oauth.extension.find(e => e.url === 'token').valueUri;
+    } catch (e2) {
+      // not sure what to do if both fail?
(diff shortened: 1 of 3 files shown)


src/utils/client.js
```diff
@@ -35,6 +36,16 @@ async function connectToServer(url) {
     }
   };
 
+  if (secret) {
```
blangley28, line 39: So I get it because Cerner doesn't support JWT client assertion we need to add this but the SMART Backend Authorization spec says it SHALL use JWT and not client secret. I know its easier for us to adapt than make Cerner be compliant but it still rubs me the wrong way. I think at a minimum letting them know and including a comment here saying this is out of spec and only to support Cerner would help

src/utils/fhir.js
```diff
@@ -324,7 +324,10 @@ function postSubscriptionsToEHR(subscriptions) {
 
       // Create/Update Subscriptions on EHR server
       const ehrToken = await getAccessToken(ehrServer.endpoint);
-      const headers = { Authorization: `Bearer ${ehrToken}` };
+      const headers = {
+        Authorization: `Bearer ${ehrToken}`,
+        'Content-Type': 'application/fhir+json'
```
blangley28, line 329: Is it worth setting this conditionally based on the metadata? I haven't actually tested with the medmorph-fhir-server yet but I think it should work. I could imagine other servers being as picky as the subscription-proxy and requiring `application/json` instead of `application/fhir+json`.

src/utils/client.js
```diff
@@ -16,11 +16,12 @@ const error = require('../storage/logs').error('medmorph-backend:client');
 async function connectToServer(url) {
   // Generate the client_assertion jwt
   const tokenEndpoint = await getTokenEndpoint(url);
-  const clientId = servers.getServerByUrl(url).clientId;
+
+  const { clientId, secret, customScopes } = servers.getServerByUrl(url);
```
blangley28, line 20: Would you mind adding the `secret` and `customScopes` to the `servers` collection headers in `src/ui/components/Collections/Collections.jsx` so they can be edited from the ui?

blangley28 (changes requested):

src/utils/client.js
blangley28: Cerner doesn't support subscriptions, right? So is this even needed here or only in the subscriptions-proxy? Since the interface between the proxy and the sandbox is out of scope of the spec I'm totally fine with having that use client secret

dehall: That's a good point. This is a reference implementation of a specification. Cerner doesn't claim to support the MedMorph spec or the SMART backend auth spec so adding code here to support them is strange. Cerner and Epic don't support either the Subscription resource, nor any approach to build a subscription capability on top of what they do support, so I'm not sure if it's worth using the subscription proxy at all. (They require all resource queries to be by ID, so for example you can get an Encounter by ID, or all Encounters by a Patient ID, but not all Encounters.) I guess we could use it as an "auth shim"...

src/utils/fhir.js
dehall: Can we reliably pick it from metadata? The Epic open sandbox metadata (https://fhir.epic.com/interconnect-fhir-oauth/api/FHIR/R4/metadata) includes `json` (alongside `xml`) as a supported format, not `application/...`. I do see that HAPI lists the full mime type there.

blangley28 (approved):

src/utils/fhir.js
blangley28: So from the CapabilityStatement resource it looks like the correct syntax is just `json` or `xml`, which doesn't help us determine whether or not to use `fhir+`

danlee1025 (approved):
Just needs a rebase.

PR #45, opened 2021-05-03
Add env variables to remove auth

danlee1025:
- Added `REQUIRE_AUTH` and `REQUIRE_AUTH_FOR_OUTGOING` environment variables.
- If set to `false`, the app will authorize all requests and not retrieve access tokens for outgoing requests.

### DIFF
--- a/.env.dev
+++ b/.env.dev
@@ -5,3 +5,5 @@ AUTH_TOKEN_URL=http://moonshot-dev.mitre.org:8090/auth/realms/backend_app/protoc
 DATA_TRUST_SERVICE=http://localhost:3005
 BASE_URL=http://localhost:3000
 ADMIN_TOKEN=admin
+REQUIRE_AUTH=false
+REQUIRE_AUTH_FOR_OUTGOING=false
--- a/.env.dev
+++ b/.env.dev
@@ -5,5 +5,5 @@ AUTH_TOKEN_URL=http://moonshot-dev.mitre.org:8090/auth/realms/backend_app/protoc
 DATA_TRUST_SERVICE=http://localhost:3005
 BASE_URL=http://localhost:3000
 ADMIN_TOKEN=admin
-REQUIRE_AUTH=false
-REQUIRE_AUTH_FOR_OUTGOING=false
+REQUIRE_AUTH=true
+REQUIRE_AUTH_FOR_OUTGOING=true
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ Certain configuration properties MUST be set via enviornment variables. The easi
 | DATA_TRUST_SERVICE | Yes | The base url for the data/trust service server. |
 | BASE_URL | Yes | The base url for this server. |
 | ADMIN_TOKEN | No | The value of the admin token to bypass authorization. If unset no admin token can be used. |
+| AUTH | No | If false or unset, an access token is not required to make requests.
+| AUTH_REQUIRED_FOR_OUTGOING | No | If false or unset, the backend app will not acquire an access token before sending requests to other servers.
 
 ```env
 # Port number for the server, defaults to 3000 if not provided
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -42,6 +42,8 @@ Certain configuration properties MUST be set via enviornment variables. The easi
 | DATA_TRUST_SERVICE | Yes | The base url for the data/trust service server. |
 | BASE_URL | Yes | The base url for this server. |
 | ADMIN_TOKEN | No | The value of the admin token to bypass authorization. If unset no admin token can be used. |
+| AUTH | No | If false or unset, an access token is not required to make requests.
```
blangley28, line 45: See comment in other repo but I still think the default (unset) should be to use security (especially since it is part of the spec)

src/utils/client.js
```diff
@@ -49,6 +49,9 @@ async function connectToServer(url) {
  * @param {string} url - the base url of the server to connect to
  */
 async function getAccessToken(url) {
+  if (!process.env.REQUIRE_AUTH_FOR_OUTGOING || process.env.REQUIRE_AUTH_FOR_OUTGOING === 'false')
+    return '';
```
blangley28, line 53: Just out of curiosity (as an open to question to anyone on the team reading this): the app will still send an Authorization header just as `Authorization: Bearer `. That shouldn't cause any problems right? I know on our servers we won't even look at the Authorization header but could it be an issue else where?

blangley28 (commented):
This was much simpler than I thought for some reason. Looks like it covers everything though

src/utils/client.js
danlee1025: yea I started to remove the headers from all the requests but realized just sending the `Authorization: Bearer ` would be easier since all the requests call `getAccessToken`. I didn't see any problems while testing with our servers. I would imagine it should be fine with other servers too. If the server were to check the token it would just extract an empty string.

README.md
```diff
@@ -42,6 +42,8 @@ Certain configuration properties MUST be set via enviornment variables. The easi
 | DATA_TRUST_SERVICE | Yes | The base url for the data/trust service server. |
 | BASE_URL | Yes | The base url for this server. |
 | ADMIN_TOKEN | No | The value of the admin token to bypass authorization. If unset no admin token can be used. |
+| AUTH | No | If false, an access token will not be required to make requests. Defaults to true if unset. |
+| AUTH_REQUIRED_FOR_OUTGOING | No | If false, the backend app will not acquire an access token before sending requests to other servers. Defaults to true if unset. |
```
blangley28, line 46: Do we think it is worth making this a server by server thing? I don't think it would be hard but I'm not sure if that's overkill. I assume everyone at this connectathon will be doing no auth since that's what Dragon said but if someone was ready it would be fun to test it out

blangley28 (approved):
LGTM. My comments are more for discussion but I'm happy with the way it is now

src/utils/client.js
dehall: It's impossible to say what arbitrary servers might do, but since the goal here is to disable the authorization process altogether I think it's safe to assume the target endpoint won't be looking at the Authorization header at all

README.md
dehall: If major EHRs are participating, I wonder if they even have the ability to turn off auth. We could do that but I think it's probably overkill since it will only be relevant for this sort of connectathon event, where the other participants don't have everything implemented yet.

.env.dev
```diff
@@ -4,4 +4,6 @@ AUTH_CERTS_URL=http://moonshot-dev.mitre.org:8090/auth/realms/backend_app/protoc
 AUTH_TOKEN_URL=http://moonshot-dev.mitre.org:8090/auth/realms/backend_app/protocol/openid-connect/token
 DATA_TRUST_SERVICE=http://localhost:3005
 BASE_URL=http://localhost:3000
-ADMIN_TOKEN=admin
\ No newline at end of file
+ADMIN_TOKEN=admin
+REQUIRE_AUTH=false
```
dehall, line 8: Two quick things - can you make the default here `true` and add these environment variables to `docker-compose.yml` and `docker/aws-docker-compose.yml` files?

dehall (approved):

PR #49, opened 2021-05-11
Config UI

KeeyanGhoreshi:
Adds a collection for the config.  There's no additional UI made for this, since the config is a collection and is being displayed as a regular collection.  The only difference is that you're not allowed to delete config items.  This was built on top of the config branch, so you should be able to see the default contents of the config file in the collection.

### DIFF
--- a/src/ui/components/Collections/Collections.jsx
+++ b/src/ui/components/Collections/Collections.jsx
@@ -101,8 +101,8 @@ const Collections = props => {
           { value: 'secret', label: 'secret', edit: true },
           // NOTE: secret is only used for client_secret_basic auth.
           // see src/utils/client.js : connectToServer for more info
-          { value: 'customScopes', label: 'customScopes', edit: true },
-        //   { value: 'type', label: 'type', edit: false }
+          { value: 'customScopes', label: 'customScopes', edit: true }
+          //   { value: 'type', label: 'type', edit: false }
         ];
         return {
           headers,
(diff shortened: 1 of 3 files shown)


src/ui/components/Collections/Collections.jsx
```diff
@@ -101,9 +101,8 @@ const Collections = props => {
           { value: 'secret', label: 'secret', edit: true },
           // NOTE: secret is only used for client_secret_basic auth.
           // see src/utils/client.js : connectToServer for more info
-          { value: 'customScopes', label: 'customScopes', edit: true },
-          { value: 'type', label: 'type', edit: false }
-        ];
+          { value: 'customScopes', label: 'customScopes', edit: true }
+]        ];
```
blangley28, line 105: Extra `]` here
```suggestion
        ];
```

blangley28 (commented):
Besides that one syntax error this works well. Functionally it does everything we need to but it would be nice if it looked more like a settings menu (i.e. toggles for the true/false values like this https://material-ui.com/components/switches/) instead of just the generic collections screen. 
I know we don't have designs but we could come up with something simple. @dehall what do you think? Or is the ability to edit them enough?

src/ui/components/Collections/Collections.jsx
KeeyanGhoreshi: i wonder why this didn't produce an error and the linter didn't say anything

dehall:
Agreed a more targeted UI would be nice, but I think the ability to edit them is enough for now. Maybe Noranda has some existing designs for settings pages, or we can ask her for something new, and we can make that a new task. This might be useful during the connectathon so I don't want to hold it up too long

PR #50, opened 2021-05-11
Fetch resources in create-report action

dehall:
Fetches resources as specified by the `PlanDefinition.action.input` field on the `create-report` action. Note that this should work when the trigger is a Patient or an Encounter, not sure about other resource types such as Condition yet. 

I used the attached encounter resource, pointing to all our internal FHIR servers, for testing:
[mitre_sample_enc.json.txt](https://github.com/mcode/medmorph-backend/files/6462135/mitre_sample_enc.json.txt)

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -67,8 +67,83 @@ const baseIgActions = {
     // Through the source client?
     return context;
   },
-  'create-report': context => {
+  'create-report': async context => {
     // the output might define a profile
+    const action = context.action;
+
+    if (action.input) {
+      // input field defines which resource types to include
+
+      for (const input of action.input) {
+        const resourceType = input.type;
+        // ASSUMPTION: only types that are linked to a patient will be included here
+        // so for example Substance resources, which represent a Thing rather than an activity
+        // should not be here
+        const params = [];
+
+        // worth making this a switch instead of if?
+        if (resourceType === 'Patient') {
+          if (context.patient) {
+            // we already added the patient resource directly
+            // (see reporting_workflow:initializeContext)
+            continue;
+          } else {
+            // assume we have context.encounter
+            const patientId = context.encounter.subject.reference;
+            const patient = (await readFromEHR(patientId)).data;
+            context.records.push(patient);
+            // skip the rest of the logic below
+            continue;
+          }
+        } else if (resourceType === 'Encounter') {
+          if (context.encounter) {
+            // we already added the encounter resource directly
+            // (see reporting_workflow:initializeContext)
+            // ASSUMPTION: if given an encounter context,
+            //   we only care about that one, not all encounters
+            continue;
+          } else {
+            // assume we have context.patient
+            params.push(`subject=Patient/${context.patient.id}`);
+          }
+        } else {
+          // TODO: logic to check which resources actually have these fields?
+          // will be necessary if we get away from the simple US core types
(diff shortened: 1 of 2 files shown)


src/utils/actions.js
```diff
@@ -66,8 +66,83 @@ const baseIgActions = {
...
+            // ASSUMPTION: if given an encounter context,
+            //   we only care about that one, not all encounters
+            continue;
+          } else {
+            // assume we have context.patient
+            params.push(`subject=Patient/${context.patient.id}`);
+          }
+        } else {
+          // TODO: logic to check which resources actually have these fields?
+          // will be necessary if we get away from the simple US core types
```
blangley28, line 110: This is a good point. For the connectathon it looks like this is fine but definitely something we should look into (and possibly create a new task for the backlog if required)

src/utils/actions.js
```diff
@@ -66,8 +66,83 @@ const baseIgActions = {
...
+        //   // TODO: what if the server doesn't include profiles or support the search filter?
+        //   params.push(`_profile=${input.profile[0]}`);
+        //   // note also specifying profile may not be ideal
+
+        // eslint-disable-next-line max-len
+        //   // see: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Requiring.20meta.2Eprofile.20be.20populated
+        // }
+
+        const query = resourceType + '?' + params.join('&');
+        const bundle = (await readFromEHR(query)).data;
```
blangley28, line 137: Instead of using `await` in a loop, keep a list of promises then and then call `.all`

blangley28 (changes requested):

src/utils/actions.js
dehall: Created MEDMORPH-121 for this

src/utils/reporting_workflow.js
```diff
@@ -275,6 +275,11 @@ async function executeWorkflow(context) {
     const action = planDef.action.find(a => a.id === currentActionId);
     context.action = action;
 
+    if (action.action) {
+      // connectathon hack - merge everything from the nested action into this one
+      Object.assign(action, action.action[0]);
+    }
+
```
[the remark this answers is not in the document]
dehall: Agree in spirit, but I don't want to be caught unprepared if for any reason the un-fixed PlanDef is still the one they use. I'd rather delete this once the change is published. And in the meantime, this won't break anything else that doesn't use a nested action

blangley28 (approved):
LGTM, testing with an Observation and it was successfully added to the content bundle. One quick thought - we only add to the report to the completed reports table on the "complete-report" action. For the connectathon it would be nice to have this populated. We could either:
1. Insert it on submit and complete reporting
2. Ask Dragon to add "complete-reporting" action to the PlanDefinition. I'm not sure what other people do during this step but it kind of sounded like every PlanDefinition should end with this anyway.

dehall:
Yeah I noticed that too. Based on the list of action codes it would seem like that one should always be last, but it's not specifically called out as necessary anywhere: https://build.fhir.org/ig/HL7/fhir-medmorph/provisioning.html 

It's a few more clicks, but for something like the connectathon where we probably won't have dozens of reports, I'm fine with just clicking through the Reporting table to find the submitted bundle

danlee1025 (approved):

PR #55, opened 2021-05-17
normalize urls when compared

KeeyanGhoreshi:
This introduces the `normalizeUrl` library to the backend.  Any instance of URLs being compared has been replaced with the `compareUrls` function which normalizes the urls before they're compared.  This should ensure consistency.  The normalize function does produce an error when it's fed a bad URL, and I was considering adding a try/catch, but there shouldn't be a situation (outside of the `storeRequests` function) where it's fed a bad url.

### DIFF
--- a/src/storage/logs.js
+++ b/src/storage/logs.js
@@ -2,6 +2,7 @@ const db = require('./DataAccess');
 const { LOGS, ERRORS, REQUESTS, NOTIFICATIONS } = require('./collections');
 const debugConsole = require('debug');
 const { v4: uuidv4 } = require('uuid');
+const normalizeUrl = require('normalize-url');
 
 function debug(location) {
   return message => {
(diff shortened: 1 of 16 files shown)


src/storage/logs.js
```diff
@@ -45,10 +46,18 @@ function storeRequest(request) {
   if (modifiedBody.password) {
     delete modifiedBody.password;
   }
+  let url = '';
+  try {
+    // some request urls can't be normalized and
+    // throw an error
```
blangley28, line 52: Which ones? Is this a potential issue for compareUrl?

blangley28 (approved):
Looks good. I don't the the extension urls needs to normalized but that shouldn't cause any issues

src/storage/logs.js
KeeyanGhoreshi: requests to the medmorph backend itself come in as relative, and the normalize function fails on "urls" like `/index`.  We don't use compareUrl anywhere this would be an issue, though.

dehall (approved):

PR #58, opened 2021-05-26
Refactor FHIR Collections

blangley28:
Refactor FHIR collections to move resources to its own property. Since this touches so many different pieces I will need help stress testing it. I started testing a bunch of different workflows and caught some errors but I wouldn't be surprised if I missed one or two places.

### DIFF
--- a/src/handlers/crudHandler.js
+++ b/src/handlers/crudHandler.js
@@ -46,13 +46,13 @@ function updateHandler(collectionName, req, res) {
 }
 
 function deleteHandler(collectionName, req, res) {
-  let resource;
+  let entryToDelete;
   if (req.query.id) {
-    resource = db.select(collectionName, r => r.id === req.query.id)[0];
+    entryToDelete = db.select(collectionName, r => r.id === req.query.id)[0];
     db.delete(collectionName, r => r.id === req.query.id);
   } else if (req.query.fullUrl) {
     const fullUrl = base64url.decode(req.query.fullUrl);
-    resource = db.select(collectionName, r => compareUrl(r.fullUrl, fullUrl))[0];
+    entryToDelete = db.select(collectionName, r => compareUrl(r.fullUrl, fullUrl))[0];
     db.delete(collectionName, r => compareUrl(r.fullUrl, fullUrl));
   } else {
     res.send('Must include id or fullUrl').status(StatusCodes.BAD_REQUEST);
@@ -61,9 +61,9 @@ function deleteHandler(collectionName, req, res) {
 
   // Also delete subscription from EHR
   if (collectionName === SUBSCRIPTIONS) {
-    deleteSubscriptionFromEHR(resource.resource);
+    deleteSubscriptionFromEHR(entryToDelete.resource);
   } else if (collectionName === PLANDEFINITIONS) {
-    const subscriptions = getSubscriptionsFromPlanDef(resource.fullUrl);
+    const subscriptions = getSubscriptionsFromPlanDef(entryToDelete.fullUrl);
 
     // Delete all subscriptions associated with plan definition
     subscriptions.forEach(subscription => {
--- a/src/handlers/crudHandler.js
+++ b/src/handlers/crudHandler.js
@@ -61,13 +61,13 @@ function deleteHandler(collectionName, req, res) {
 
   // Also delete subscription from EHR
   if (collectionName === SUBSCRIPTIONS) {
-    deleteSubscriptionFromEHR(resource);
+    deleteSubscriptionFromEHR(resource.resource);
   } else if (collectionName === PLANDEFINITIONS) {
     const subscriptions = getSubscriptionsFromPlanDef(resource.fullUrl);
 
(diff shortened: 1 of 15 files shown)


KeeyanGhoreshi (commented):
Trying to edit the fullUrl of PlanDefinition throws `TypeError: Cannot read property 'id' of undefined`, and continues to error and crash after reloading and trying to view the PlanDefinition collection in the UI

src/ui/components/Collections/CollectionRow.jsx
```diff
@@ -44,6 +47,7 @@ function CollectionRow(props) {
     // the event is a custom object returned from
     // JSONInput, not a regular event
     if (event.jsObject) {
+      console.log(event.jsObject);
```
dehall, line 50: Stray console.log here

src/handlers/crudHandler.js
```diff
@@ -61,13 +61,13 @@ function deleteHandler(collectionName, req, res) {
 
   // Also delete subscription from EHR
   if (collectionName === SUBSCRIPTIONS) {
-    deleteSubscriptionFromEHR(resource);
+    deleteSubscriptionFromEHR(resource.resource);
```
dehall, line 64: Maybe the variable name shouldn't be resource if it's not actually a FHIR resource. I'd propose `entry` or `resourceWrapper` or just `deletedItem` because that's how it's being used, a link back to the item that was deleted.

src/ui/components/Collections/CollectionRow.jsx
```diff
@@ -27,11 +28,13 @@ function CollectionRow(props) {
...
     headers.forEach(header => {
       if (!header.viewOnly) {
-        bundle[header.value] = state[header.value];
+        const value = state[header.value];
+        if (value) bundle[header.value] = value;
       }
     });
 
-    axios.put(`/collection/${selectedCollection}?id=${bundle.id}`, bundle).then(() => {
+    const query = bundle.id ? `id=${bundle.id}` : `fullUrl=${base64url(bundle.fullUrl)}`;
```
dehall, line 36: I'm not clear on why this changed. Are there cases where we don't have an id?

KeeyanGhoreshi (changes requested):
~~When you change the criteria on a subscription, in the JSON editor, it modifies the state such that the internal `resource` field on the object gets updated, but the entirety of the `resource` object's fields get pulled up into the top level.~~

~~The display prefers the top-level version, so it doesn't update visually.~~

src/ui/components/Collections/CollectionRow.jsx
blangley28: I don't think it was correct before. IDs are not unique for our FHIR collections. The backend crud handler supported fullUrl query params but the UI wasn't using them

KeeyanGhoreshi:
Ignore my previous comment, my db had old un-refactored resources in it

src/ui/components/Collections/Collections.jsx
```diff
@@ -143,7 +143,9 @@ const Collections = props => {
         headers = [
           { value: 'id', label: 'id', edit: false },
           { value: 'fullUrl', label: 'fullUrl', edit: true },
-          { value: 'criteria', label: 'criteria', edit: true },
+          { value: 'criteria', label: 'criteria', edit: false },
```
KeeyanGhoreshi, line 146: `edit: false` doesn't mean you can't define it when you make the resource.  If you click "add new" and make a new subscription, you can define the criteria but it gets put at the top level, instead of in the `resource`.  This would be a `viewOnly` field like the server icon:

`{ value: 'icon', label: '', edit: false, viewOnly: true }`

src/ui/components/Collections/Collections.jsx
```diff
@@ -143,7 +143,9 @@ const Collections = props => {
...
+          { value: 'server', label: 'location', edit: false },
+          { value: 'timestamp', label: 'last updated', edit: false },
```
KeeyanGhoreshi, line 148: same problem with timestamp as above, when you make a new resource this field should be automatically filled out.

KeeyanGhoreshi (changes requested):

src/ui/components/Collections/Collections.jsx
blangley28: Also added to PlanDefinition `name`

src/handlers/crudHandler.js
blangley28: renamed to `entryToDelete`

KeeyanGhoreshi:
This isn't necessarily something this PR needs to fix, but if you create a new subscription in the UI, don't include a `fullUrl`, and save it, then attempt to restart the backend, it will fail.

blangley28:
Update the headers to add an optional `required` param. This way the UI checks and will not persist the update if a required field is empty. Shows the error in a small snackbar alert.

KeeyanGhoreshi (approved):

dehall (approved):

PR #61, opened 2021-06-03
Dynamic client registration

blangley28:
Attempt to register the client with the PHA using [OAuth 2.0 Dynamic Client Registration Spec](https://datatracker.ietf.org/doc/html/rfc7591). 

One big note about this: the spec is pretty loose on the client metadata. Keycloak only support OpenID which further restricts the OAuth 2.0 spec and does not work for the `client_credentials` grant flow required for SMART Backend Authorization. I set up an Okta authorization server to test this but:
1. It requires an extra metadata not defined in the spec called `application_type`. This must be set to `service` for the client to be registered as defined in the SMART Backend Auth spec. This extra information should not break anything else so I left it in there for now.
2. It does not support `jwks_uri`, only `jwks`. The SMART Backend Auth spec allows for `jwks` but `jwks_uri` is preferred as it allows keys to be rotated. Unfortunately there does not seem to be a way to determine which claims on the metadata an authorization server supports. 
3. It requires an access token on the registration endpoint. It seems like all of the identity management platforms require this, and thus are mostly intended for a developer to register an application and not a piece of software to register itself. 

Open to suggestions on what this PR should and should not include plus discussion on our own authorization server.

### DIFF
--- a/config.js
+++ b/config.js
@@ -32,7 +32,7 @@ const configInit = [
     description: 'Whether the BSA will fetch and use a token when making external requests'
   },
   {
-    id: configVars.DYANMIC_CLIENT_REGISTRATION,
+    id: configVars.DYNAMIC_CLIENT_REGISTRATION,
     value: false,
     display: 'Dynamic Client Registration',
     description: 'Whether the BSA will attempt to dynamically register with PHAs'
--- a/config.js
+++ b/config.js
@@ -2,14 +2,16 @@ const configVars = {
   ADMIN_TOKEN: 'admintoken',
   DATA_TRUST_SERVICE: 'datatrustservice',
   REQUIRE_AUTH: 'requireauth',
-  REQUIRE_AUTH_FOR_OUTGOING: 'requireauthout'
+  REQUIRE_AUTH_FOR_OUTGOING: 'requireauthout',
+  DYANMIC_CLIENT_REGISTRATION: 'dynamicclientregistration'
 };
 const configInit = [
   {
     id: configVars.ADMIN_TOKEN,
     value: 'admin',
     display: 'Admin token',
-    description: 'Allows incoming requests to use a fixed bearer token for testing the interactions, should be unset in production.'
+    description:
+      'Allows incoming requests to use a fixed bearer token for testing the interactions, should be unset in production.'
   },
   {
     id: configVars.DATA_TRUST_SERVICE,
@@ -28,6 +30,12 @@ const configInit = [
     value: true,
     display: 'Require auth out',
     description: 'Whether the BSA will fetch and use a token when making external requests'
+  },
+  {
+    id: configVars.DYANMIC_CLIENT_REGISTRATION,
+    value: false,
+    display: 'Dynamic Client Registration',
+    description: 'Whether the BSA will attempt to dynamically register with PHAs'
   }
 ];
 
--- a/config.js
+++ b/config.js
@@ -3,7 +3,7 @@ const configVars = {
   DATA_TRUST_SERVICE: 'datatrustservice',
   REQUIRE_AUTH: 'requireauth',
   REQUIRE_AUTH_FOR_OUTGOING: 'requireauthout',
-  DYANMIC_CLIENT_REGISTRATION: 'dynamicclientregistration'
+  DYNAMIC_CLIENT_REGISTRATION: 'dynamicclientregistration'
 };
 const configInit = [
   {
--- a/src/utils/client.js
+++ b/src/utils/client.js
@@ -109,6 +109,11 @@ async function registerServer(url) {
   };
 
   const registrationEndpoint = await getRegistrationEndpoint(url);
+  if (!registrationEndpoint) {
+    error(`Unable to register new server ${url}: dynamic registration not supported`);
+    return null;
+  }
+
   return axios
     .post(registrationEndpoint, metadata)
     .then(result => {
@@ -158,7 +163,7 @@ async function getTokenEndpoint(url) {
  * Get the registration_endpoint from the .well-known/smart-configuration
  *
  * @param {string} url - the fhir base url
- * @returns registration_endpoint
+ * @returns registration_endpoint or null if not found (not supported)
  */
 async function getRegistrationEndpoint(url) {
   try {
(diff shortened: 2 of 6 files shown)


dehall (commented):
> It requires an access token on the registration endpoint. It seems like all of the identity management platforms require this, and thus are mostly intended for a developer to register an application and not a piece of software to register itself.

I'm not seeing this in the code you added, am I just blind?

Other than that the code is pretty clean and straightforward, a lot simpler than I realized. Though that limitation of the token pretty much means it can't work in practice. One thing we might want to consider is a config setting to allow/disable attempting to do dynamic registration. Other than that, I think we could come up with a lot of "nice to haves" around the idea of dynamic registration, things like a collection for reports that weren't able to be submitted or the server wasn't able to be registered, that would be submitted later; or trying to register early, like as soon as we load a PlanDef we try to register with the server. But those wouldn't be part of this task.

blangley28:
> I'm not seeing this in the code you added, am I just blind?

Not blind, I removed it since it required my specific admin auth token. Considering Okta will only show you the hash of it after it is created I figured it's pretty powerful and did not want it pushed to the repo. If you want to test we can talk out of band.

> Though that limitation of the token pretty much means it can't work in practice. 

Agreed, I haven't worked with this kind of registration before so I think when other implementers are considering authentication it is worth raising for discussion at a working group call or connectathon.

> One thing we might want to consider is a config setting to allow/disable attempting to do dynamic registration.

At first I thought about this but why would you really want to turn it off? Because you know it will never work and want to stop filling the logs/notifications with failed attempts?

> I think we could come up with a lot of "nice to haves" around the idea of dynamic registration, things like a collection for reports that weren't able to be submitted or the server wasn't able to be registered, that would be submitted later; or trying to register early, like as soon as we load a PlanDef we try to register with the server. But those wouldn't be part of this task.

I've had thoughts about both of these two. I think more effort should be put into notifying the user of when a report fails. The notifications/error tab is helpful for us but can be difficult to understand which reports failed. I will create a new task for this. 
As for registering early I think that's small enough (and crucial enough) to implement here. Assuming the BSA is not using TTP (because then it is their responsibility to register with the PHAs), the PHA should really be registered as soon as the PlanDefinition is received and thus notifying the user before it has actually been submitted. This should be just moving some code around.

dehall:
> At first I thought about this but why would you really want to turn it off? Because you know it will never work and want to stop filling the logs/notifications with failed attempts?

Yeah exactly. That way we have the code here but it doesn't actually get called. I don't love the idea of merging in code that we know is going to error out on a regular basis.

src/utils/knowledgeartifacts.js
```diff
@@ -55,6 +57,9 @@ function fetchEndpoint(url, planDefinition) {
     if (endpoint) {
       db.upsert(ENDPOINTS, endpoint, r => compareUrl(r.fullUrl, endpoint.fullUrl));
       debug(`Endpoint/${endpoint.id} saved to db`);
+
+      const baseUrl = endpoint.address.split('/$process-message')[0];
```
blangley28, line 61: Specifically looking for feedback to see if this is suffice

dehall: Only thing I would add is a check to only try to register if the server isn't already in the servers collection. (or if it doesn't already have a client_id or something). I don't see that in the registerServer function but I think it would be fine either right here in fetchEndpoint or there in registerServer

config.js
```diff
@@ -2,7 +2,8 @@ const configVars = {
   ADMIN_TOKEN: 'admintoken',
   DATA_TRUST_SERVICE: 'datatrustservice',
   REQUIRE_AUTH: 'requireauth',
-  REQUIRE_AUTH_FOR_OUTGOING: 'requireauthout'
+  REQUIRE_AUTH_FOR_OUTGOING: 'requireauthout',
+  DYANMIC_CLIENT_REGISTRATION: 'dynamicclientregistration'
```
dehall, line 6: typo on the field name, though the value and functions that use it elsewhere are good

src/utils/client.js
```diff
@@ -82,9 +84,54 @@ async function getAccessToken(url) {
...
+/**
+ * Register the backend app with the server and add it to the servers collection
+ *
+ * @param {string} url - the server base url
+ * @returns the server object if successful, otherwise null
+ */
+async function registerServer(url) {
+  if (
+    configUtil.getDynamicClientRegistration() &&
+    configUtil.getDynamicClientRegistration() === 'false'
```
KeeyanGhoreshi, line 104: Wouldn't this work without the `configUtil.getDynamicClientRegistration()`?  If it was null, the equality comparison would still return false.

KeeyanGhoreshi (approved):

src/utils/knowledgeartifacts.js
```diff
@@ -55,6 +58,9 @@ function fetchEndpoint(url, planDefinition) {
...
```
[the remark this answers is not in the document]
blangley28: So by default `getReferencedResource` will only return the resource but in this scenario we need the fullUrl as well. As such I had previously updated the method to take in an extra parameter called `returnFullUrl`. When this is true it returns the data structure `{ fullUrl: string, resource: Resource }` instead of just `Resource`. 

The fix was just to reference the `.resource` property in this `fetchEndpoint` function. This also fixes the problem where we were logging `Endpoint/undefined` by changing the line above to be `endpoint.resource.id`.

src/utils/client.js
```diff
@@ -81,9 +83,50 @@ async function getAccessToken(url) {
...
+}
+
+/**
+ * Register the backend app with the server and add it to the servers collection
+ *
+ * @param {string} url - the server base url
+ * @returns the server object if successful, otherwise null
+ */
+async function registerServer(url) {
+  if (configUtil.getDynamicClientRegistration() === 'false') return null;
```
dehall, line 101: After the rebase this is now a boolean so 
```suggestion
  if (!configUtil.getDynamicClientRegistration()) return null;
```

config.js
```diff
@@ -28,6 +30,12 @@ const configInit = [
     value: true,
     display: 'Require auth out',
     description: 'Whether the BSA will fetch and use a token when making external requests'
+  },
+  {
+    id: configVars.DYANMIC_CLIENT_REGISTRATION,
```
dehall, line 35: Typo came back

dehall (commented):
Sorry, a couple tiny last items here. Then it should be good to merge

dehall (approved):

PR #62, opened 2021-06-04
add config ui

KeeyanGhoreshi:
You can go to the config tab to look at the UI.  The designs didn't have a "submit" button anywhere or a save button, so the text fields just send an update request when you blur the text field.  I'm open to suggestions for other solutions to updating the text fields.

### DIFF
--- a/config.js
+++ b/config.js
@@ -7,19 +7,23 @@ const configVars = {
 const configInit = [
   {
     id: configVars.ADMIN_TOKEN,
-    value: 'admin'
+    value: 'admin',
+    display: 'Admin token'
   },
   {
     id: configVars.DATA_TRUST_SERVICE,
-    value: 'http://localhost:3005'
+    value: 'http://localhost:3005',
+    display: 'Data trust service'
   },
   {
     id: configVars.REQUIRE_AUTH,
-    value: 'true'
+    value: 'true',
+    display: 'Require auth'
   },
   {
     id: configVars.REQUIRE_AUTH_FOR_OUTGOING,
-    value: 'true'
+    value: 'true',
+    display: 'Require auth out'
   }
 ];
 
--- a/src/ui/components/Config/ConfigRow.jsx
+++ b/src/ui/components/Config/ConfigRow.jsx
@@ -1,0 +1,94 @@
+import React, { useEffect, useMemo, useState } from 'react';
+import axios from 'axios';
+import { useQueryClient } from 'react-query';
+import TextField from '@material-ui/core/TextField';
+import Switch from '@material-ui/core/Switch';
+
+import PropTypes from 'prop-types';
+import useStyles from './styles';
+
+const ConfigRow = props => {
+  const classes = useStyles();
+  const queryClient = useQueryClient();
+  const { option } = props;
+  const [value, setValue] = useState(option.value);
+  const [changed, setChanged] = useState(false);
+
+  const isBool = useMemo(() => {
+    let val = option.value;
+    val = val === 'true' ? true : val === 'false' ? false : val;
+    return typeof val === 'boolean';
+  }, [option]);
+
+  const handleChange = event => {
+    setValue(event.target.value);
+    setChanged(true);
+  };
+
+  const handleUpdateBool = event => {
+    updateOption(event.target.checked);
+  };
+
+  const handleUpdateText = () => {
+    if (changed) {
+      updateOption(value);
+      setChanged(false);
+    }
+  };
+
+  const updateOption = optionValue => {
+    const result = { ...option, value: optionValue };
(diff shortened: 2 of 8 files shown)


src/ui/components/Dashboard/DashboardLayout.jsx
```diff
@@ -32,7 +33,7 @@ function DashboardLayout() {
         label: `Notifications ${newNotifs ? `(${unviewedNotifs.length})` : ''}`,
         component: <Notifications notifs={unviewedNotifs} />
       },
-      { key: 'config', label: 'Config', component: <Collections selectedCollection="config" /> },
+      { key: 'config', label: 'Config', component: <Config config="config" /> },
```
dehall, line 36: it is certainly fun to have 1 line with the word "config" in it 5 times, but the component takes no props so we should delete those last 2: `config="config"`

webpack.config.js
```diff
@@ -7,6 +7,8 @@ module.exports = {
     filename: 'bundle.js'
   },
   watch: watch,
+  devtool: "#eval-source-map",
```
dehall, line 10: What does this do? I'm not opposed to adding it if it does something good, but I'm not clear on what exactly it does from https://webpack.js.org/configuration/devtool/ ... and if the # changes anything

config.js
```diff
@@ -7,19 +7,23 @@ const configVars = {
 const configInit = [
   {
     id: configVars.ADMIN_TOKEN,
-    value: 'admin'
+    value: 'admin',
+    display: 'Admin token'
```
dehall, line 11: Can these `display` be made a little more descriptive? Admin token is used to allow incoming requests to use a fixed bearer token for testing the interactions, should be unset in production. Data trust service is the base endpoint of the data trust service. Require auth is whether or not authorization is required to access endpoints on the BSA. Require auth out is whether the BSA will fetch and use a token when making external requests. Feel free to rework the text or designs as needed for space reasons

blangley28: Perhaps a display and also a helper tooltip with a little extra text to explain what the value is/used for?

src/ui/components/Config/ConfigRow.jsx
```diff
@@ -0,0 +1,94 @@
...
+const ConfigRow = props => {
+  const classes = useStyles();
+  const queryClient = useQueryClient();
+  const { option } = props;
+  const [value, setValue] = useState(option.value);
+  const [changed, setChanged] = useState(false);
+
+  const isBool = useMemo(() => {
+    let val = option.value;
+    val = val === 'true' ? true : val === 'false' ? false : val;
```
blangley28, line 19: Is it easier to do this than just change the physical variable type to be a boolean?

src/ui/components/Config/ConfigRow.jsx
```diff
@@ -0,0 +1,94 @@
...
+    return typeof val === 'boolean';
+  }, [option]);
+
+  const handleChange = event => {
```
blangley28, line 23: All of these functions should use `useCallback`

blangley28 (changes requested):
Played around with the textui and it always seemed to update correctly for me so I think you can leave that as it is. One additional thing to implement for the task which I don't see here is clicking on the Settings in the side panel should open up the config tab.

src/ui/components/Config/ConfigRow.jsx
blangley28: As a side note: it seems like once I edit the value it saves to the database as the boolean type anyway. So I think editing this directly in the initial configVars should work. Just make sure the places before which were checking the `configUtil.getRequireAuth() === 'false'` and `configUtil.getRequireAuthOutgoing() === 'false'` are updated accordingly

webpack.config.js
KeeyanGhoreshi: This makes errors and console logs output from the file/line they come from in the source code instead of the compiled webpack bundle.   Makes debugging a lot easier.  The hashtag is supposed to help but I guess it might be obsolete?

https://stackoverflow.com/questions/43564604/what-does-the-pound-sign-means-in-webpacks-devtool-option

src/ui/components/Dashboard/Menu.jsx
```diff
@@ -14,15 +14,19 @@ import { useQueryClient } from 'react-query';
 function Menu(props) {
   const classes = useStyles();
   const queryClient = useQueryClient();
-  const { open, callback } = props;
+  const { open, setContentKey, callback } = props;
 
+  const setKey = useCallback(() => {
```
blangley28, line 19: Since this function always sets the contentKey to config can it be renamed to something more descriptive?

blangley28 (approved):
LGTM

dehall (approved):

PR #63, opened 2021-06-09
CQL Support

danlee1025:
- Support for cql in plandefinition check trigger codes condition
- Pulled in valueset and library loading functionality from STEAM abstraction tool
- Executes ELM and checks if condition is included in result

Testing:
Manually trigger `plandefinition-cancer-example-cql`. I added Patient `pat02` which fails the trigger check. The CQL expression is just checking if the patient's age is >= 40.

Use this encounter to pass the trigger check
```
{
  "resourceType": "Encounter",
  "id": "encounter-manual-trigger-example",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Encounter with patient pat01</div>"
  },
  "status": "in-progress",
  "class": {
    "system": "http://terminology.hl7.org/CodeSystem/v3-ActCode",
    "code": "IMP",
    "display": "inpatient encounter"
  },
  "subject": {
    "reference": "Patient/pat01"
  }
}
```

Use this encounter to fail the trigger check
```
{
  "resourceType": "Encounter",
  "id": "encounter-manual-trigger-example",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">Encounter with patient pat02</div>"
  },
  "status": "in-progress",
  "class": {
    "system": "http://terminology.hl7.org/CodeSystem/v3-ActCode",
    "code": "IMP",
    "display": "inpatient encounter"
  },
  "subject": {
    "reference": "Patient/pat02"
  }
}
```


src/utils/cql/evaluateCql.js
```diff
@@ -0,0 +1,65 @@
+const { Client } = require('cql-translation-service-client');
+const { Library, Executor, Repository } = require('cql-execution');
+const { PatientSource } = require('cql-exec-fhir');
+const { Base64 } = require('js-base64');
```
dehall, line 4: This is pretty minor but I noticed it right away - rather than add a new dependency, is there any reason we can't use `btoa` / `atob` here?

src/utils/cql/evaluateCql.js
```diff
@@ -0,0 +1,65 @@
...
+const client = new Client(
+  'http://moonshot-dev.mitre.org:8080/cql/translator?annotations=true&result-types=true'
+);
+
+async function evaluateCQL(resources, expression, library, patientId) {
+  const elm = await fetchELM(library);
+  const valueSetLoader = new ValueSetLoader(library);
+  const valueSetMap = await valueSetLoader.seedValueSets();
+  const result = executeElm(resources, elm, valueSetMap);
+  return result.patientResults[patientId][expression] === 1;
```
dehall, line 18: This should be `=== true` since we expect the expressions to be boolean-valued. 
```suggestion
  return result.patientResults[patientId][expression] === true;
```

src/storage/collections.js
```diff
@@ -11,5 +11,6 @@ module.exports = {
   MESSAGES: 'messages',
   REQUESTS: 'requests',
   NOTIFICATIONS: 'notifications',
-  CONFIG: 'config'
+  CONFIG: 'config',
+  LIBRARIES: 'libraries'
```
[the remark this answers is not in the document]
blangley28: Just to clarify this is the first _fhir_ collection which isn't resourceType + s. The places this happen are:

- `routes/index` `/trigger` POST handler: this shouldn't matter though because the resource here is from a trigger and is thus not a Library
- `services/subscription` `handleReportTriggerResource`: this shouldn't matter either because the resource here is also from a trigger and is thus not a Library
- `utils/actions` `extract-research-data`: this could _potentially_ cause an issue if for some reason a Library was requested as part of report
- `utils/fhir` `getResources`: this is probably the only real potential issue. If the Library is returned back with the request for the PlanDefinition then this code could be run for a Library resource.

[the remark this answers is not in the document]
danlee1025: I did not really think about the consequences when I added this collection. Should I make it `Librarys` to remain consistent?

src/utils/actions.js
```diff
@@ -428,10 +436,17 @@ function createBundle(records, type) {
   return bundle;
 }
 
-function evaluateExpression(expression, resources, variables = {}) {
+function evaluateExpression(expression, resources, variables = {}, library, patientId) {
```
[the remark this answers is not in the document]
danlee1025: I actually removed the function since the pieces of logic are basically one liners and performed the language check in `check-trigger-codes`. Let me know if you still think it would be better to break them up into separate functions.

src/storage/collections.js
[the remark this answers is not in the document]
dehall: Sounds good. There are some other options, but all have their own downsides, so I think this is probably the best. 

- we could make the constant LIBRARIES but make the collection name librarys. This could get confusing if you try to go directly into the DB and can't find "libraries"
- we could make a more complex "pluralize" function, or import a library. Probably overkill
- we could patch the english language but i think that PR is unlikely to be merged

src/utils/cql/evaluateCql.js
```diff
@@ -0,0 +1,62 @@
+const { Client } = require('cql-translation-service-client');
+const { Library, Executor, Repository } = require('cql-execution');
+const { PatientSource } = require('cql-exec-fhir');
+const atob = require('atob');
+const axios = require('axios');
+const { ValueSetLoader } = require('./ValueSetLoader');
+const configUtil = require('../../storage/configUtil');
+
+async function evaluateCQL(resources, expression, library, patientId) {
+  const elm = await fetchELM(library);
```
blangley28, line 10: So this works but just as a thought we could convert the CQL to ELM when the library is added to the database. And then do a check here to see if the elm exists

src/utils/actions.js
```diff
@@ -82,19 +84,37 @@ const baseIgActions = {
...
+
+          if (language === 'text/fhirpath') {
+            return isTrue(fhirpath.evaluate(resources, expression, variables));
+          } else if (language === 'text/cql') {
+            // Spec allows for multiple libraries, how can we tell which one is the main?
+            // Intentionally simplified and assuming the first library is the main
+            let library;
+            const planDefLib = context.planDefinition.library;
+            if (planDefLib?.length > 0) {
+              library = db.select(LIBRARYS, l => l.resource.id === planDefLib[0])[0].resource;
```
blangley28, line 100: Should we add another check here that `db.select(...)[0]` is defined? If there was an error adding the library to the database or a user deleted the library from the collection then `db.select(...)[0]` will return undefined and could cause an exception

src/utils/knowledgeartifacts.js
```diff
@@ -130,6 +131,44 @@ async function fetchValueSets(url, planDefinition, bundle) {
...
+ * @param {string} url - fhir server base url to fetch from
+ * @param {PlanDefinition} planDefinition - PlanDefinition to search for referenced Library
+ * @param {Bundle} bundle - the Bundle the PlanDefinition was returned in,
+ *                          in case it already contains the Library we want
+ */
+async function fetchLibraries(url, plandefinition, bundle) {
+  const { library: libraryIds } = plandefinition;
+  if (!libraryIds || libraryIds.length === 0) return;
+
+  for (const id of libraryIds) {
```
blangley28, line 146: Are these always ids? I thought they were supposed to be canonical urls

danlee1025: Hmm the library I used was exported from pathways-builder and just had the id in the library field. Should I change it to reference the url field?

src/utils/cql/evaluateCql.js
danlee1025: Added this.

src/utils/actions.js
danlee1025: Makes sense, I added a check here and return false if the library is not in the database.

src/utils/cql/evaluateCql.js
blangley28: One additional pieces - if `library.elm` is not set here then when it is converted it could be updated in the database. We expect it to always have the elm so perhaps this is overkill (for instance the elm isn't set yet because it's still being converted so there's no point trying again)

src/utils/actions.js
blangley28: I noticed here if `library` is undefined then the function returns false. Should that also be the behavior when `planDefLib.length <= 0`? Under that scenario `library` would still be undefined but the call to evaluate still happens.

danlee1025: Good catch, I moved the `library` check after the `if` statement.

src/utils/cql/evaluateCql.js
danlee1025: Possibly overkill but I added the database upsert to the `fetchElm` function.

src/utils/actions.js
```diff
@@ -82,19 +84,44 @@ const baseIgActions = {
...
+              debug(`Library not defined in PlanDefinition or not found in database.`);
+              return false;
+            }
+
+            return evaluateCQL(
+              createBundle(resources, 'content'),
+              expression,
+              library,
+              context.patient.id
+            );
```
blangley28, line 114: Unless we care about the debug on line 105 this can be simplified to
```suggestion
            const planDefLib = context.planDefinition.library;
            if (planDefLib?.length > 0) {
              const libraryId = planDefLib[0];
              const library = db.select(LIBRARYS, l => l.resource.id === libraryId)[0];
              if (library) {
                 return evaluateCQL(
                  createBundle(resources, 'content'),
                  expression,
                  library,
                  context.patient.id
                );
              }
            }
            return false;
```

src/utils/cql/evaluateCql.js
```diff
@@ -0,0 +1,75 @@
...
+    elm = response.data;
+  } else if (contentInfoTranslate && contentInfoTranslate.url) {
+    const response = await axios.get(contentInfoTranslate.url);
+    elm = await client.convertBasicCQL(response.data);
+  } else {
+    throw new Error('Could not fetch ELM off of Library resource');
+  }
+
+  // Add converted ELM to database
+  db.upsert(LIBRARYS, { ...library, elm }, l => l.resource.id === resource.id);
```
blangley28, line 45: This should use fullUrl instead of resource.id as the key

blangley28 (approved):
LGTM

src/utils/cql/evaluateCql.js
```diff
@@ -0,0 +1,75 @@
...
+const { LIBRARYS } = require('../../storage/collections');
+const debug = require('../../storage/logs').debug('medmorph-backend:evaluateCql');
+const db = require('../../storage/DataAccess');
+const configUtil = require('../../storage/configUtil');
+
+async function evaluateCQL(resources, expression, library, patientId) {
+  const elm = library.elm || (await fetchELM(library));
+  const valueSetLoader = new ValueSetLoader(library);
+  const valueSetMap = await valueSetLoader.seedValueSets();
+  const result = executeElm(resources, elm, valueSetMap);
```
dehall, line 16: We lost the sample PlanDef on the internal FHIR server, so testing is a little tough at the moment, but is this where `valueSetMap` should be used? In `executeElm` below the third param is for dependent libraries, so like if we were using FHIRHelpers or something it would go in there

danlee1025: I pushed the library back up. I left out a small piece in `executeElm` that I just pushed up. This seems to be where the `abstraction-tool` used the `valueSetMap`

dehall (approved):

PR #64, opened 2021-06-21
Add mitre docker-compose file

blangley28:
The CD on pathways.mitre.org for the backend does not work because the docker-compose file has the baseurl set to localhost. This adds a new docker-compose file with updated env like the aws one in the same directory. This file is identical to the `src/docker-compose.yml` except for changing `localhost` to `pathways.mitre.org`

### DIFF
--- a/docker/mitre-docker-compose.yml
+++ b/docker/mitre-docker-compose.yml
@@ -1,0 +1,18 @@
+version: "3"
+
+services:
+  medmorph_backend:
+    image: medmorph_backend
+    container_name: medmorph_backend
+    restart: on-failure
+    ports:
+      - "3000:3000"
+    environment:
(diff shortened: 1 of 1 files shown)


danlee1025 (approved):
LGTM, did we want to use the https URL?

blangley28:
> did we want to use the https URL?

Yea good call. What are the mappings for those two ports? I'm looking through the nginx config but don't see the proxy settings

dehall (approved):
I'm ok with the http urls as-is. I'll leave it unmerged in case you want to switch to https though

danlee1025:
> > did we want to use the https URL?
> 
> Yea good call. What are the mappings for those two ports? I'm looking through the nginx config but don't see the proxy settings

8770 maps to 3000 if you wanted to use the https url.

